PR #1, opened 2019-10-17
Add python_requires 2.7 or 3.6+ to setup.py

matthewfeickert:
This will make it clear to PyPI what Python requires are needed and also make it clear to the user when they try to install it.

This PR also adds classifier metadata for PyPI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,14 @@ setup(
     long_description=open('README.md').read(),
     long_description_content_type="text/markdown",
     url="https://github.com/andrzejnovak/mplhep/",
+    classifiers=[
+        "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+    ],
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Given what @lgray had mentioned at the coffee break, I force pushed a change that also allows support for Python 2.7

andrzejnovak:
@matthewfeickert Not necessarily. Is there a good case not to? I thought it was pretty much a python standard.

matthewfeickert:
> Not necessarily. Is there a good case not to? I thought it was pretty much a python standard.

Yes, but a lot of people also think that 79 chars is too restrictive. pyhf formats with Black, and there [we use 88 lines](https://github.com/diana-hep/pyhf/blob/master/pyproject.toml#L2) as that seems to be a more reasonable length and it what the original author of Black recommends.

matthewfeickert:
@andrzejnovak I think we've broken all of the CI by merging this in now though. We'll have to open up another PR that changes the line length that signals an error.

matthewfeickert:
Ah no! I see that you are steps ahead of me and have already done this in https://github.com/andrzejnovak/mplhep/commit/521e160b13261089352114cc650958307c2a498c. :) Thanks!

PR #11, opened 2019-11-19
Add notebooks to tests

matthewfeickert:
Use [papermill](https://github.com/nteract/papermill) to add testing of Jupyter notebooks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,8 +10,8 @@ INSTALL_REQUIRES = [
 ]
 
 extras_require = {
-    'test': ['pytest', 'pytest-mpl'],
-    'develop': ['flake8', 'twine'],
+    'test': ['pytest', 'pytest-mpl', 'papermill~=1.0', 'nteract-scrapbook~=0.3'],
+    'develop': ['flake8', 'jupyter', 'twine'],
 }
 extras_require['complete'] = sorted(set(sum(extras_require.values(), [])))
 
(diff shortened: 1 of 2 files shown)


matthewfeickert:
@andrzejnovak This is ready for review.

andrzejnovak:
Let me test how this works, because I haven't seen it before. But I like the idea of example notebooks also being tests.

matthewfeickert:
> Let me test how this works, because I haven't seen it before. But I like the idea of example notebooks also being tests.

Sure. [Papermill](https://github.com/nteract/papermill) is kinda fantstic as it allows you to execute a Jupyter notebook from outside the kernel and with [scrapbook](https://github.com/nteract/scrapbook) you can actually collect output from the notebook that papermill is executing. It is really nice.

We on the pyhf team us it to [test all of our example notebooks](https://github.com/diana-hep/pyhf/blob/4df46d60267e14075e15b503e9285eea444a8ebc/tests/test_notebooks.py) and @betatim of Jupyter/Binder has a [nice talk that mentions papermill from Swiss Python Summit 2018](https://youtu.be/USb6ju7ApSs).

andrzejnovak:
@matthewfeickert So is the point of this just making sure notebooks are executable top-down?

matthewfeickert:
> So is the point of this just making sure notebooks are executable top-down?

Yes, though more importantly, if there is a PR that introduces API changes this also catches if any of the examples in the notebook would break as a result. So it gives you confidence that your examples will _always_ be up to date, as they must be by definition as you executed them.

andrzejnovak:
Ok, I understand better now. Thanks.

PR #15, opened 2019-11-22
Scale yerr when density=True

marinang:
Fix an issue that the error bars were huge when calling

```python
h, bins = np.histogram(data, bins=40, range=(0, 8))
mplhep.histplot(h, bins=bins, histtype="errorbar", yerr=True, density=True)
```

Now the _yerr is scaled like h in the code. I have also duplicated one test with normalised histogram.


andrzejnovak:
@marinang Thanks for catching that. Could you fix the flake8 error?

marinang:
@andrzejnovak  done :D

andrzejnovak:
@marinang Published to pip

PR #35, opened 2020-01-16
ci: Add release via bumpversion tag GHA workflow

matthewfeickert:
When a PR is labeled with one of the following by (and only by &mdash; if other devs need to be given publishing power their usernames can be added to this file) @andrzejnovak:

- `bumpversion/major`
- `bumpversion/minor`
- `bumpversion/patch`

this workflow will trigger. This will run `bumpversion` corresponding to the `patch`/`minor`/`major` of the label, update the commit to refer to the issue number filed, and update the tag to summarize the commits since the previous tag to the new tag.

The changes will be commit messages that follow the semantic format `$semantic: $description`.

When the PR is merged in, then the commit and tag will be pushed to master as well (using a [GitHub personal access token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) named ` GITHUB_PAT`) which will trigger the publish release.

All of this workflow is thanks to @kratsg who implemented this first in [pyhf PR 650](https://github.com/scikit-hep/pyhf/pull/650).

PR #36, opened 2020-01-16
Revert "ci: Add release via bumpversion tag GHA workflow"

andrzejnovak:
Reverts scikit-hep/mplhep#35


matthewfeickert:
@andrzejnovak  This doesn't actually need to get reverted. It only affects PRs that have the bumpversion labels on them.

PR #37, opened 2020-01-16
dummy

andrzejnovak:
Test actions?


matthewfeickert:
Let me check pyhf. If things aren't failing there then maybe remove the `GITHUB_PAT` for the time being? Permissions might be wrong there.

PR #87, opened 2020-02-21
Remove duplicate `ls`, `linestyle` arguments

marinang:
In order to avoid the deprecation warning:

``` python
/usr/local/lib/python3.7/site-packages/mplhep/plot.py:256: MatplotlibDeprecationWarning: Saw kwargs ['ls', 'linestyle'] which are all aliases for 'linestyle'.  Kept value from 'linestyle'.  Passing multiple aliases for the same property will raise a TypeError in 3.3.
  label=_labels[0], **kwargs)
```

Because in `errorbar` both are used. 

Edit: I removes in line 255 the argument `ls` because it is already defined in `err_defaults` however for the the second part

``` python
...
if _nh == 1:
    ax.errorbar(_bin_centers, h, yerr=_yerr,
                        label=_labels[0], **kwargs)
else:
    for i in range(_nh):
        ax.errorbar(_bin_centers, h[i], yerr=_yerr[i], linestyle='none',
                            label=_labels[0], **_chunked_kwargs[i])
...
```
I let the `linestyle` argument.

### DIFF
--- a/mplhep/plot.py
+++ b/mplhep/plot.py
@@ -252,7 +252,7 @@ def histplot(h, bins, weights=None, yerr=None, variances=None,
             if k not in kwargs.keys():
                 kwargs[k] = v
         if _nh == 1:
-            ax.errorbar(_bin_centers, h, yerr=_yerr, linestyle='none',
+            ax.errorbar(_bin_centers, h, yerr=_yerr,
                         label=_labels[0], **kwargs)
         else:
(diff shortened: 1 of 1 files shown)


andrzejnovak:
Aren't we still passing linestyle twice like this?

marinang:
At least I don't get the warning anymore on my side so I don't think so.

PR #89, opened 2020-02-27
Add LHCb style sheet

graipher:
Here is a style-sheet that tries to replicate the plotting style used by the LHCb experiment.
It is not faithful in all regards (notably colours and markers), but comes quite close.

The option `legend.title_fontsize` only works in recent versions of `matplotlib` and throws a warning otherwise.
The style uses LaTex by default, with the Times New Roman font (if installed).

Not sure how much work it is to convert it to the Python version you have for ATLAS and CMS.

### DIFF
--- a/mplhep/styles_lhcb.py
+++ b/mplhep/styles_lhcb.py
@@ -1,0 +1,86 @@
+from cycler import cycler
+
+colors = ['#1f77b4', '#ff7f0e', '#2ca02c', '#d62728', '#9467bd',
+          '#8c564b', '#e377c2', '#7f7f7f', '#bcbd22', '#17becf']
+markers = ['o', 's', 'D', '^', 'v', '<', '>', 'P', 'X', '*']
+
+ROOT = {
+    # Plot properties
+    "axes.labelsize": 32,
+    "axes.linewidth": 2,
+    "axes.facecolor": "white",
+    # Custom colors
+    "axes.prop_cycle": cycler('color', colors) + cycler('marker', markers),
+    "axes.formatter.min_exponent": 3,
+    # Figure properties
+    "figure.figsize": (12, 9),
+    "figure.dpi": 100,
+    # Outer frame color
+    "figure.facecolor": "white",
+    "figure.autolayout": True,
+
+    # Set default font to Times New Roman
+    "font.family": "serif",
+    "font.serif": ["Times New Roman"],
+    "font.size": 14,
+    "font.weight": 400,
+
+    # Draw the legend on a solid background
+    "legend.frameon": True,
+    "legend.fancybox": False,
+    # Inherit the background color from the plot
+    "legend.facecolor": "inherit",
+    "legend.numpoints": 1,
+    "legend.labelspacing": 0.2,
+    "legend.fontsize": 28,
+    "legend.title_fontsize": 28,
+    # Automatically choose the best location
+    "legend.loc": "best",
+    # Space between the handles and their labels
+    "legend.handletextpad": 0.75,
+    # Space between the borders of the plot and the legend
+    "legend.borderaxespad": 1.0,
+    "legend.edgecolor": "white",
+
+    # Lines settings
+    "lines.linewidth": 4,
+    "lines.markeredgewidth": 0,
+    "lines.markersize": 8,
+
+    # Saved figure settings
+    "savefig.bbox": "tight",
+    "savefig.pad_inches": 0.1,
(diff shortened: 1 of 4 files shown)


andrzejnovak:
Welcome @graipher, I was hoping someone from LHCb would show up. The pythonization is easy as they stylesheet can be just translated to a dictionary like https://github.com/scikit-hep/mplhep/blob/master/mplhep/styles_cms.py. Feel free to add it to your PR or if not I'll get to it at some point. 

Also python 2 and old matplotlib are deprecated. Though I have been backporting compatible stuff for now.

mplhep/styles_lhcb.py
```diff
@@ -0,0 +1,81 @@
+
+LHCb = {
+    # Plot properties
+    "axes.labelsize": 32,
+    "axes.linewidth": 2,
+    "axes.facecolor": "white",
+    # Custom colors
+    # axes.prop_cycle: cycler('color', ["1f77b4", "ff7f0e", "2ca02c", "d62728", "9467bd", "8c564b", "e377c2", "7f7f7f", "bcbd22", "17becf"]) + cycler('marker', ['o', 's', 'D', '^', 'v', '<', '>', 'P', 'X', '*'])
```
andrzejnovak: why is this uncommented?

andrzejnovak:
Flake is not happy 
```
mplhep/styles_lhcb.py:8:89: E501 line too long (211 > 88 characters)
# Many lines of:
mplhep/styles_lhcb.py:55:23: E241 multiple spaces after ':'
```

mplhep/styles_lhcb.py
graipher: Because I didn't get around to testing it yet ;)

graipher:
Ah, I thought my linter would fix that, will fix it manually.

graipher:
@andrzejnovak Please have a look now. Can't quite test it on my system, because `packaging` does not want to install, so the import fails. 
I'm also not sure if I tied it in correctly into the global package structure and gave it appropriate names

andrzejnovak:
Looks good to me. Let me check it out locally first. I don't think I have tests implemented in the CI to check everything here properly.

andrzejnovak:
So I would consider naming it (or aliasing it `LHCb` and adding it to `styles` as well.

I ran a small plot test:
```
import mplhep as hep
import numpy as np
plt.style.use(hep.styles_lhcb.ROOT)
data = np.random.normal(0, 1, 10000)
plt.hist(data)
hep.lhcb.text("Simulation")
```

Can you confirm this is the look you are after? It looks to me the top label is kind of small.

image

graipher:
In general it looks OK, but I agree the text is probably not right yet. As noted in the original stylesheet, the font size should be 28 (I guess it currently uses the 14 set for all text).

In addition, there seems to be a problem with the glyph for `-`, since it is just boxes.

graipher:
Also, if you find a way to always enforce 

```
plt.xlabel(..., ha='center', y=1)
plt.ylabel(..., ha='right', x=1, y=1)
plt.minorticks_on()
```

I would be quite glad :)

andrzejnovak:
For the unicode minus glyph, it's fine if you also include the usetex flag (though I am not sure how latex solves it, since the char is missing from the font sadly).
You could enforce it to always usetex I guess, though it makes rendering slower.

For the labels unfortunately matplotlib currently doesn't let you set it by default. I made a PR about this few months back, but it will only arrive with matplotlib 3.3 most likely. For now I use a "magic" function to do these minor style adjustments https://github.com/scikit-hep/mplhep/blob/master/mplhep/plot.py#L481  . We could have a dedicated lhcb version of that.

andrzejnovak:
Btw if you want to keep the mail address out of the log you could alwyas squash the pr

andrzejnovak:
@graipher  Is this good to merge or do you want to put `usetex` as part of the nominal style? Without tex I don't think there's a way to properly render a unicode minus in Times New Roman

andrzejnovak:
@graipher Ah found it actually, you want to add `'axes.unicode_minus' : False,` and then the no TeX option should be fine as well

andrzejnovak:
Ok, automated publishing is having problems, but other than that this looks good to me. Thanks a lot for the PR :)

andrzejnovak:
@graipher https://pypi.org/project/mplhep/0.1.6/ should have it

graipher:
@andrzejnovak Nice, and thanks for the help on the way!

PR #93, opened 2020-03-03
feat: Sync styles, add style tests

andrzejnovak:
- Rename LHCb/CMS experiment styles to their real name
- Alias CMS to ROOT
- Show warnings when calling `<exp>.style.ROOT`
- Add style tests
  - LHCb label is buggy


andrzejnovak:
@graipher Could you take a look at this PR? It renames the `ROOT` style to `LHCb`, so that all the experiments get the same interface.

graipher:
@andrzejnovak Looks good, that's what I called it initially as well, but then I saw that it was called `ROOT` in `cms`, so I renamed it :smile:.

PR #95, opened 2020-03-06
blacked

HDembinski:
This patch applies "black" formatting by the Python black module. It is easier to work with autoformatted code and "black" is becoming a standard in Scikit-HEP.

### DIFF
--- a/mplhep/__init__.py
+++ b/mplhep/__init__.py
@@ -1,4 +1,5 @@
 import os
+
 # Import counter
 import requests as req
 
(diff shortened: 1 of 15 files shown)


andrzejnovak:
I am not super sold on black. I see the benefit of not having to think about formatting, but kind of like the flexibility of flake. Anything else to consider apart from all the cool kids are doing it?

HDembinski:
It makes it easier for everyone to contribute. black is already used by most Scikit-HEP devs. So it is easier for us to submit patches if the code is already blacked. "all the cool kids are using it" is a valid argument in this case. See comments on black on https://gitter.im/Scikit-HEP/community

andrzejnovak:
Ok, I took another look and while I am not happy about some of the choices I guess that is the way of black. I was mostly looking for some arguments along the lines of http://www.locallyoptimal.com/blog/2019/08/23/why-you-should-use-black-for-your-python-style-linting/

If you want to fix this up to agree with the rest of the CI I'll merge it, or I'll do it myself in the upcoming days.

PR #99, opened 2020-03-08
Allow hist objects to go into histplot(), Use edges by default, small fixes

### LINKED ISSUE #98: Wrong default display for mplhep.histplot
The command
`hep.histplot(h, bins, ax=axs[0])`
does not produce a histogram in HEP-style. It draws markers at the edges, which is wrong, and the lines of the histogram do not go down to zero.

`hep.histplot(h, bins, edges=True, ax=axs[1])`
is closer to the expected skyline, but the markers are still wrong. Apart from the wrong markers, this should be the default. In fact, I cannot imagine that one ever wants to have `edges=False`.

andrzejnovak:
Alright, seems the consensus is leaning this way. Closes #98


mplhep/plot.py
```diff
@@ -20,7 +20,7 @@
 
 def histplot(h, bins, weights=None, yerr=None, variances=None,
```
HDembinski: It would be ideal if you could change the names of the functions and the defaults at the same time. In other words, keep the interface of histplot as it is, but deprecate it. Add a new function `hist` without the `edges` keyword, because we always want the edges. We never want to draw without them.

HDembinski: In fact, the interface of `hist` should get a few more changes. Interfaces should be minimally sufficient. Redundancies must be avoided.

*  This means `variances` should be removed, because it seems redundant with `yerr`. 
* `weights`should be removed, it makes no sense for prebinned histograms
* Perhaps: Remove `stack` and automatically draw a stacked histogram when h is a collection of arrays
* Remove `densitymode`, just use `density` like `pyplot.hist` does (if the user wants the other kind of density, which is not a real density anyway so the name is wrong, they have to calculate it themselves)
* Remove `binticks`, there are already ways to manipulate ticks in mpl

henryiii: I was hoping to work on this interface as part of Hist (probably during the Summer with a student, but maybe a bit before). I would like the new interface to support histogram objects (boost-histograms and hist histograms), probably by making edges optional, and pulling them from the first object if not given. In that case, just keeping the current interface patched up and "good enough" and the replacing it with a new one later might be the best course of action.

Are you supporting Python 2? If not, you could also make the new interface keyword only after bins, which would reduce the potential for errors when switching and would be safer to change in the future. `hist(h, bins=None, *, ...)`

henryiii:
Does this also fix the odd markers shown in the plot in the issue?

andrzejnovak:
@henryiii py2 is deprecated, I'm backporting simple fixes and occasional feature that's simple enough.

Agree about making it take histogram object from numpy and boost histogram. 

@HDembinski we should quite simply not redefine well established mpl functions. I'm playing with a prototype that monkeyclasses Axes, so one could then do
`import mplhep.pyplot as plt` and then have the histplot callable on an axis. We can then then patch labels for positions as well until mpl 3.3.3 happens. I'll rename it, but not to hist. 

This whole package is a convenience wrapper, you can already do everything in matplotlib.
- variance is / will be useful for plotted objects where the errors should be automatic, yerr gives direct access
- i agree about weights, can't think of a valid use case
- stack / densitymode: sometimes you want to different things, you might want all series normalized, or on top of each other, or normalized together. These are every day analysis things that one shouldn't have to implement manually each time
- binticks: see starter comment, it's convenient and aligning ticks with edges is aesthetically pleasing

mplhep/plot.py
HDembinski: I want the new interface as soon as possible, so I can tell the people in our group that they can plot numpy histograms now. Agreed, the interface should work with numpy histograms and boost-histogram. So then the first argument should either be the numpy histogram tuple, or a boost-histogram or whatever. No bins argument.
`hist(h, *, ...)` with h either boost-histogram or `(n, edges)` from numpy.histogram.

andrzejnovak: I can accommodate the numpy hist very quickly. In fact, in way it kind of already works `hep.histplot(*np.histogram())` I haven't played with boost::histogram yet, but addin a object check and extracting values/edges/variances should be straightforward.

henryiii: The nice thing in @HDembinski's suggestion is that histograms are already objects; you can't add a boost-histogram with `*h`, since that doesn't expand to a numpy histogram (you can with `*h.to_numpy()` of course, but that's not ideal). Then you don't need to figure out what to do with the bin argument. I think this is sort of how Aghast and Uproot handle numpy histograms; they are always passed around and recognized as a tuple. @jpivarski could correct me if I'm wrong.

HDembinski:
@andrzejnovak Please use the reply mode on the individual review items instead of commenting on the PR. Also, your defensiveness towards my constructive criticism is not very productive. I am a very experienced interface designer, so you can assume that my advice is pretty solid.

* variance: can you elaborate what you mean? I still think this has to go.
* stack: I said "perhaps", so I am not 100% sure, but you can already plot several histograms individually, simply by calling the function several times. When you pass all histogram at once to the function, it makes no sense to do the same thing. Instead we could then plot them stacked and loose the superfluous stack keyword.
* densitymode: this has to become just "density", nothing else makes sense
*

mplhep/plot.py
henryiii: @HDembinski I think it's okay to wait a bit (if we do need to) to make sure we get the interface right. If @andrzejnovak likes your suggestion of h being a tuple for numpy, that sounds okay to me.

HDembinski:
In case you don't know, I designed Boost.Histogram and contributed many patches to Boost
https://github.com/boostorg/histogram
I am the maintainer of iminuit, pyhepmc, I co-maintain boost-histogram with Henry, and I contributed significant patches to matplotlib. I just don't make up my mind based on some subjective aesthetics, but I read books about design.

mplhep/plot.py
andrzejnovak: I always intended that this should take objects in the end, so I am happy with that change. I do think it should be possible to plot a hist manually without constructing a tuple or an object first.
So I would prefer to solve this with `hist(h, bins=None, *, ...)` or `hist(h, *, bins=None, ...)` if need be, but imho one should be able to call `hist([1,2,1], [0,1,2,3])` rather than `hist([1,2,1], bins=[0,1,2,3])` Keeping in mind I think this should be somewhat compatible to usual matplotlib functions and `plt.hist` allows for equivalent use.

@HDembinski would you find this acceptable?

andrzejnovak: One concern I have about taking objects is that any library this accommodates is then a dependency, which is imho kind of a weird position. I almost like the idea better of forcing any object through aghast and checking if one can extract the info by calling `h/edges/variances` on it.

Maybe there's a better solution I don't know about.

henryiii: This library should not depend on any histogram libraries. Hist may depend on this one, but not the other way around! What you can do, though, is to specify an interface. It could be something like `.axes.edges`, or something like `__edges__` (it's technically disallowed to use custom double underscore names, though). If the attribute is there, use it; if it's not, then treat it like a normal array.

andrzejnovak:
> @andrzejnovak Please use the reply mode on the individual review items instead of commenting on the PR. Also, your defensiveness towards my constructive criticism is not very productive. I am a very experienced interface designer, so you can assume that my advice is pretty solid.
> 
> * variance: can you elaborate what you mean? I still think this has to go.

Use-case for variances is for example with numpy or coffea histograms, which don't calculate errorbars, so the user would have to do it himself. This is a simple functionality that mplhep can easily help the user with. Use-case for errors is one might want a histogram-like plot where the error bars are something else e.g. ratio.
> * stack: I said "perhaps", so I am not 100% sure, but you can already plot several histograms individually, simply by calling the function several times. When you pass all histogram at once to the function, it makes no sense to do the same thing. Instead, we could then plot them stacked and lose the superfluous stack keyword.

You could probably convince me of this because by changing the default mode to just stacking everything, no real functionality would be lost. Then again is it really necessary to force the user to write the loop explicitly themselves? Also as I mentioned elsewhere I think there's value in keeping in line with the matplotlib interface and `plt.hist` has that option
> * densitymode: this has to become just "density", nothing else makes sense

densitymode allows for the choice for calculating density for the stack or separately for each hist, I've personally made plots of both types. But I agree this wouldn't be necessary if the stacking behaviour was made default.
> * ~ticks aligned with bin edges is the expected behavior~ actually no, this function should not draw ticks, the Locator is responsible for placing the ticks; if people want something different, they can mess with the ticks themselves

I would like that functionality easily accessible because I use it often. I am happy to factor it out if you have a better suggestion, but the only alternative I can think of is a new function that would adjust it taking and returning the axes object, which is again not great
> One big newby mistake is to put functionality in some interface that can be easily obtained by other means. We want precise and sharp tools for one specific job, and not unwieldly swiss-army knifes.


@HDembinski I am sorry you've found my responses frustrating. I'm very glad you brought in a number of suggestions, but I may need some back and forth on them as I have use cases for almost all the elements I included. 

This library is primarily supposed to save user time. Sacrificing convenience in the name of simplicity does not help with that.

mplhep/plot.py
HDembinski: You need to depend on boost-histogram to support boost-histograms, since Python is so dynamic. For example, to support boost-histogram, you could check whether the method `.to_numpy()` exists and call it to get the usual numpy.histogram tuple.

@andrzejnovak Since you opened another PR with the monkey-patching idea (which was useful to understand better what were you want to go, thanks), I added a more detailed discussion on the two options that I see for the interface there. https://github.com/scikit-hep/mplhep/pull/100

HDembinski: > @HDembinski I think it's okay to wait a bit (if we do need to) to make sure we get the interface right. If @andrzejnovak likes your suggestion of h being a tuple for numpy, that sounds okay to me.

From the Zen of Python:
"Now is better than never.
Although never is often better than *right* now."

I propose we discuss and implement the new interface in #100, which is new, so anything we do there does not have to be backward compatible with the existing `histplot` function.

HDembinski:
> Use-case for variances is for example with numpy or coffea histograms, which don't calculate errorbars, so the user would have to do it himself. This is a simple functionality that mplhep can easily help the user with. Use-case for errors is one might want a histogram-like plot where the error bars are something else e.g. ratio, so you want to calculate it yourself and set it directly.

I think that this function should not draw error bars at all. This then simplifies the interface a lot.

* In publication quality plots, histograms are drawn **either** as points with errorbars or as a skyline. In high-quality plots, we never draw a skyline with errorbars for numerous reasons that I do not want to list here.
* We already have `errorbar` to draw points with error bars. People should use that.

If we decide to nevertheless use this function also to histograms as points with errorbars, then we still need to fix the interface. I am against the `variance` keyword for many reasons.

It is a bad name, the variance is the square of the standard deviation. What we draw is an error bar, which may or may not be the standard deviation. In general it is a confidence interval which only coincides with +/- standard deviation for normal distributed data, and histogram counts are not normal distributed.

It is redundant with `yerrs`.

You cannot automatically compute errorbars for numpy histograms, boost-histograms, or coffea histograms, because you cannot predict how the user wants to draw them and you do not know what the data in the histogram means. 1) There are multiple ways to compute the errorbars from the histogrammed information, sqrt(N) is only one of several options, and 2) You do not know whether the histogram contains counts at all, because people may have already scaled the histogram by some factor and then sqrt(N) is wrong. If the histogram was build from weighted events (very common in HEP), then you cannot compute the error as sqrt(N).

So would end with several other keywords to customize the error bar calculation. `variance` has to go. If we want to include automatic error calculation, it has to be very flexible and go through the `yerr` keyword.

* `yerr=None`: don't draw errorbars
* `yerr=<array, shape(N,)>`: draw errorbars using sigma values in array
* `yerr=<array, shape(2, N)>`: draw asymmetric errorbars with sigma-, sigma+ just like `pyplot.errorbar`
* `yerr=True`: draw errorbars according to some well-documented default method that statistical experts find suitable for Poisson-distributed data (as a statistical expert, I can suggest choice, but it won't be sqrt(N))
* `yerr=<string>`: draw errorbars according to method with name <string>, e.g. "bootstrap", "frequentist", "bayesian", (better names to be defined). See https://en.wikipedia.org/wiki/Poisson_distribution#Confidence_interval

HDembinski:
Next comment is about `stack`:

> You could probably convince me of this because by changing the default mode to just stacking everything, no real functionality would be lost. Then again is it really necessary to force the user to write the loop explicitly themselves? Also as I mentioned elsewhere I think there's value in keeping in line with the matplotlib interface and plt.hist has that option.

Writing a loop in Python is incredibly easy, but plotting multiple histograms with one call does not scale very well to complex but common use cases. Users often want to customize how each histogram is drawn (specify the color, linestyle, etc.). You can do this by allowing `color` and `linestyle` to accept sequences as well, but this lets the interface code explode for no good reason and the call becomes hard to read. I know that `pyplot.plot` allows to do make several plots at once, but this was a design mistake that they probably inherited from matlab.

HDembinski:
> @HDembinski I am sorry you've found my responses frustrating. I'm very glad you brought in a number of suggestions, but I may need some back and forth on them as I have use cases for almost all the elements I included.
This library is primarily supposed to save user time. Sacrificing convenience in the name of simplicity does not help with that.

I was trying to speed things up a bit by demonstrating my experience with interface design. We can discuss all these design decisions at length, but I can assure you that I will tell you explicitly when I am not sure about something. But there some things that I am simply sure have to be done in a different way. I am generally not interested in pushing my subjective preferences. I try very hard to take subjectiveness out of design. My design decisions are grounded on experience, insight in the needs of users and library implementers, and design principles. Having and following prinicples is very important. It reduces the potential design space and leads to consistency.

andrzejnovak:
@HDembinski Okay variances is really a misnomer here. I think I got stuck on it, after trying to extract Sumw2 from a TH1F - which is what it really should be and how it's used https://github.com/scikit-hep/mplhep/blob/master/mplhep/plot.py#L122  - should just be renamed

Most of the usage you suggest for `yerr` should be there
- [x] yerr=None: don't draw errorbars, just draw the skyline
- [x] yerr=<array, shape(N,)>: don't draw the skyline, but draw points with errorbars using sigma values in array

You can currently choose it with the skyline (default) or without by passing `histtype='errorbar`
- [x] yerr=<array, shape(2, N)>: same, but draw asymmetric errorbars with sigma-, sigma+ just like pyplot.errorbar
- [ ] yerr=True: no skyline but points, draw errorbars according to some well-documented default method that statistical experts find suitable for Poisson-distributed data (as a statistical expert, I can suggest choice, but it won't be sqrt(N))

Indeed, this currently plays the role of a shorthand/approx. An expert recommendation is welcome.
- [ ] yerr=<string>: no skyline but points, draw errorbars according to method with name , e.g. "bootstrap", "frequentist", "bayesian", (better names to be defined). See https://en.wikipedia.org/wiki/Poisson_distribution#Confidence_interval

As above

I strongly believe that whether skyline is drawn or not should be controlled by `histtype` as it is for fill vs line, because that's what it really is. Dynamically changing the plot type based on the passed values or special keyword in `yerr` is confusing.

andrzejnovak:
@HDembinski @henryiii I think to stay productive this PR should get resolved and we can continue discussing the future of the interface and interplay with `pyplot` in #100 

For now, I've addressed the following:
- Make `edges=True` default
- Force `lines.marker=''` in `step` which should rid you of the weird markers
- Deprecated use of `weights`
- Renamed `variances` to `w2` to fix what they really should be
- Enable `histplot` to take directly 
  - unwrapped tuple (a.k.a manual input)
  - numpy tuple
  - any object which has `to_numpy()` and the call results in a one 1D hist numpy tuple

This addresses stuff we found consensus on, uncontested suggestions, a bugfix and iiuc most of the requested new functionality.

Further issues and suggestions remain welcome. In particular more rigorous yerr methods @HDembinski

PR #101, opened 2020-03-11
Fix bug

LovelyBuggies:
AttributeError: 'list' object has no attribute 'T'

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ hep.histplot(h, bins)
 import numpy as np
 xbins, ybins = [0, 1, 2, 3], [0, 1, 2, 3]
 H = np.array([[2,3,2], [1,2,1], [3,1,3]])
-hep.hist2dplot(H.T, xbins, ybins)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,9 +54,10 @@ hep.histplot(h, bins)
 ```
 #### 2D Histograms
 ```
+import numpy as np
 xbins, ybins = [0, 1, 2, 3], [0, 1, 2, 3]
-H = [[2,3,2], [1,2,1], [3,1,3]]
-hep.hist2dplot(H, xbins, ybins)
+H = np.array([[2,3,2], [1,2,1], [3,1,3]])
+hep.hist2dplot(H.T, xbins, ybins)
```
eduardo-rodrigues: The previous line `np.array` is correct but the transpose is not needed.

LovelyBuggies: All right, I will remove it.

LovelyBuggies:
@eduardo-rodrigues I have removed the transpose. LGTM, and you can merge it!

PR #106, opened 2020-03-24
Errorbar: Use right labels

AndreasAlbert:
This is a tiny fix to plotting in "errorbar" mode to ensure that the right labels are used in the legend.


andrzejnovak:
Thanks a lot!

PR #111, opened 2020-04-02
fix: don't ignore binwnorm

andrzejnovak:
@AndreasAlbert Can you confirm this is expected behaviour?
```
import mplhep as hep
import numpy as np
h, bins = np.histogram(np.random.normal(10,3,1000))

f, ax = plt.subplots()
hep.histplot(h, bins, label="Nominal")
hep.histplot(h, bins, binwnorm=1, label="1")
hep.histplot(h, bins, binwnorm=0.5, label="0.5")
plt.legend()
```

image

### DIFF
--- a/mplhep/plot.py
+++ b/mplhep/plot.py
@@ -185,7 +185,7 @@ def histplot(
             binnorms = binnorms[0]
         return binnorms
 
-    if density:
+    if density or binwnorm is not None:
         density_arr = get_density(h, density, binwnorm)
         if stack and densitymode == "stack":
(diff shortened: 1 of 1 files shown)


AndreasAlbert:
Looks about right, but I can't 100% what the bin width is from the plot. If orange = blue / bin width and green = 0.5 * orange, then it's what I expect.

Thanks for fixing!

andrzejnovak:
Alright let me release this and submit a PR to bump the version in coffea

PR #121, opened 2020-04-17
delete bla.txt

HDembinski:


andrzejnovak:
I'm gonna merge this, but please use semantic commits, so that the deployment chain works properly.
Like here the commit message would be `fix: delete bla`

https://www.conventionalcommits.org/en/v1.0.0-beta.2/

It should be fine here, since I'm not gonna build a release on it.

PR #128, opened 2020-04-18
fix: backcomp mpl 3.1

### LINKED ISSUE #127: Support for matplotlib 3.1?
Is FontManager critical? i.e. could we detect mpl version on import and do the appropriate so we can still use mplhep with (slightly) older mpl

andrzejnovak:
Closes #127


andrzejnovak:
@HDembinski

PR #133, opened 2020-04-25
Allow single-histogram stacks

### LINKED ISSUE #129: Stacking a single histogram
Hi,

I am wondering about the explicit check of the number of histograms in a stack [here](https://github.com/scikit-hep/mplhep/blob/master/mplhep/plot.py#L140-L141). 

I run into this check a lot when I e.g. select datasets in my coffea histograms to be overlaid using a regular expression. In order to produce many plots in an automated way, I'd  like to be able to use the same plotting code whether my regex matches 1 or N>1 data sets.

I tried simply removing the check and the code runs fine, which is also what  I'd have expected since a single-histogram stack is well-defined. Are there any strong reasons to abort plotting in this case?

AndreasAlbert:
Closes #129

### DIFF
--- a/mplhep/plot.py
+++ b/mplhep/plot.py
@@ -137,9 +137,6 @@ def histplot(
     else:
         raise ValueError("Input not recognized")
 
-    if _nh == 1:
-        assert not stack, "Cannot stack one histogram"
-
     # Find a better way to unwrap to "real" dimentionality
     if h.ndim == 2 and len(h) == 1:  # Unwrap if [[1,2,3]]
         h = h[0]


andrzejnovak:
@AndreasAlbert Can you ammend the commit message to be compatible with https://www.conventionalcommits.org/en/v1.0.0/

AndreasAlbert:
Let's try again

andrzejnovak:
Thanks! I guess the release action gets hung up on something else. I'm gonna push a release manually.

@matthewfeickert Can you take a look at what could be tripping bump version up? I'm confused where this is coming from 
```
Run git merge --squash "origin/${GITHUB_HEAD_REF}"
merge: origin/2020-04-25_stack_single_hist - not something we can merge
```

matthewfeickert:
> Can you take a look at what could be tripping bump version up? I'm confused where this is coming from

Hey. Yeah, the way that we originally wrote this assumed that @kratsg, @lukasheinrich, and I (the core devs) would be the only people to be making releases. Looking at [the logs](https://github.com/scikit-hep/mplhep/runs/630623966#step:8:1)

```
  git merge --squash "origin/${GITHUB_HEAD_REF}"
  git commit -m "${PR_TITLE} (#${PR_NUMBER})"
  shell: /bin/bash -e {0}
  env:
    IS_MAJOR: false
    IS_MINOR: false
    IS_PATCH: true
    PR_NUMBER: 133
    PR_TITLE: Allow single-histogram stacks
    GITHUB_HEAD_REF: 2020-04-25_stack_single_hist
    BV_PART: patch
merge: origin/2020-04-25_stack_single_hist - not something we can merge
```

this makes sense as `origin/2020-04-25_stack_single_hist` **doesn't exist** &mdash; it exists only on @AndreasAlbert's fork under that name. [GitHub Actions views this as `pull/133/merge`](https://github.com/scikit-hep/mplhep/runs/630623966#step:5:384).

I don't want to dictate a release schedule to you, but this seems a bit excessive to get its own patch release. If you want to be able to make a new patch from any incoming PR then the CI is going to have to get rewritten and without having looked at it in a while I'm not sure how hard that will be to debug. It might be easier for you to create a throwaway repo to test this all out with fake releases that never deploy.

andrzejnovak:
Ah, I understand now. So the catch was that a new release can only be done from a PR, where the branch lives in origin. I guess that makes sense. 

Based on what you said, it looks like it should be easy enough fix though. Just changing https://github.com/scikit-hep/mplhep/blob/master/.github/workflows/tag.yml#L54 to instead take the PR ref, https://github.com/scikit-hep/mplhep/runs/630623966#step:5:384  That is I one can merge a ref, but I am assuming yes.

kratsg:
Yes, change it to `refs/pull/${PR_NUMBER}/merge" and it should be ok.

PR #135, opened 2020-05-12
fix: correctly plot single-hist stacks

AndreasAlbert:
This fixes a mistake I introduced in #133.

For stacks of a single histogram, the current implementation draws the cumulative of the histogram, which is not what we want.


andrzejnovak:
Thanks for the fix and sorry, about the spam - still trying to figure out releases from outside PRs.

PR #161, opened 2020-07-12
feat: support hist titles, better Uproot 4 support

henryiii:
This would support titles from the histogram libraries. I need to test locally - is there anything else I should add to test this? @andrzejnovak does this look about right?

The other thing we should support is errors from weighted histograms. That will probably be in a later PR, after we address https://github.com/scikit-hep/boost-histogram/issues/413 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: CI/CD
 
 on:
   push:
+    branches:
+      - master
+      - master-py27
   pull_request:
   # Run daily at 0:01 UTC
   schedule:
--- a/mplhep/plot.py
+++ b/mplhep/plot.py
@@ -114,7 +114,7 @@ def histplot(
 
     elif hasattr(h, "to_numpy"):
         # Generic (possibly Uproot 4)
-        _tup = h.to_numpy()
+        _tup = h.to_numpy(flow=False)
         if len(_tup) != 2:
             raise ValueError("to_numpy() method not understood")
         else:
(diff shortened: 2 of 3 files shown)


henryiii:
I'll fix the issues, but could you please consider adding pre-commit support? It's much nicer for a developer to one thing in all repositories `pre-commit run` or maybe `pre-commit install` for automatic checks) than to have to figure out what checks you are using 1 by 1. You can add any checks you want that way, and the GitHub action for pre-commit is simple. I would guess you'd put flake8 and mypy in, at least. See the scikit-hep developer pages.

andrzejnovak:
@henryiii Ok, it makes sense. I still feel a bit uneasy about trusting black, so I end up dry running it first etc.. but I see how it would simplify stuff.

henryiii:
I thought you were already running black? I saw it listed. Anyway, the point was not to add a check, just to provide a universal interface for running whatever you like.

henryiii:
Oh, maybe I misunderstood. You can easily dry run with pre-commit to see how it changes, just do `git diff` after pre-commit runs. It never changes your staging area, just your working copy. You have to add the changes manually, even in "installed" mode. I often do that.

henryiii:
@LovelyBuggies, can you take a look and see if this looks fine / works?

.github/workflows/ci.yml
```diff
@@ -2,6 +2,9 @@ name: CI/CD
 
 on:
   push:
+    branches:
```
henryiii, line 5: One not-quite-related change - the tests currently were running twice when I make a pull request from a feature branch.

andrzejnovak:
Looks good to me. You could consider adding a test using https://matplotlib.org/3.1.1/api/testing_api.html#matplotlib.testing.decorators.check_figures_equal

henryiii:
Nice to see that, will be useful. I can also check to see if the labels on axes are set. Do you mind me adding boost-histogram to the test suite?

andrzejnovak:
Yeah, that makes sense.

Issue #168, opened 2020-07-27
ATLAS style axis labels incorrect


Popinaodude:
Axis labels in ATLAS style are much too large and positions incorrect. Here are the options I made to fix:
``
    plt.style.use([hep.style.ATLAS,
                   {'axes.labelsize': 20},
                   {'axes.labelpad': 20},
                   {'axes.titlelocation': 'right'},
                   ])
``
Incidentally the  `{'axes.titlelocation': 'right'}` doesn't seem to change anything, and neither does `plt.xlabel(r"xlabel", ha='right')`. I'm not sure if that's an issue with mplhep or matplotlib

andrzejnovak:
Welcome @Popinaodude could you make a PR with the style changes? @kratsg does this look ok to you since it's your work originally?

The legend position is mpl related issue. We fixed it in next release. mplhep has a helper function for it for now  `mplhep.r_align()`

kratsg:
labelsizes are `30` by default in ATLAS (I did the conversion from the ATLAS style guide to the matplotlib). For reference

[ATLAS dictates](https://twiki.cern.ch/twiki/bin/view/AtlasProtected/PubComPlotStyle#Plot_size):
- Rectangular plot: 800x600 px
- Square plot: 600x600 px

Since the smallest dimension is 600px for plots, and the font default is 42 (font choice=4, precision=2), this means that the text size in ATLAS (set to 0.05) will be a fraction of the pad size: 600*0.05 = 30.

Popinaodude:
Perhaps it's a problem on my end, but here's the output for the same histogram at label and pad size 30 and 20 respectively:

![testplot_nofix](https://user-images.githubusercontent.com/20911987/88566063-61bb6c00-d02d-11ea-9b6b-c6a2b8b451a7.png)
![testplot_fixedstyle](https://user-images.githubusercontent.com/20911987/88566123-7b5cb380-d02d-11ea-88af-9cc55b4385c2.png)

kratsg:
@andrzejnovak is the specification in mplhep meant to be in points or pixels? If we follow atlas style, then it should be 23/24pt, or 30 pixels.

andrzejnovak:
@kratsg it should be in points. The labels do look a tad large to me, but if the current numbers are what the style guide dictates we should keep it.

@Popinaodude you've found the correct way to overwrite the default settings. If you have mpl 3.3+ there are two new rcParams
```
[...]

PR #171, opened 2020-08-11
fix: improve font sizes, padding for ATLAS style

andrzejnovak:
This should address the issue/discussion raised in #168. Typical plot would go from looking like

image

to 

image

of which the label size proportions seem more reasonable to me. 

@kratsg Popinaodude could you comment?

While we're revisiting this I would also suggest appending Tex Gyre Heros to the font list so that it doesn't fall on DejaVu Sans if none of the listed fonts is present, which is more likely to happen than in case of LHCb style which uses almost omnipresent Times.


kratsg:
Tex Gyre Heros has the wrong italic slant needed from a quick look. It definitely looks more off to me. I think if we're going to request the font to be more accessible, it should be taken up w/ the experiments directly.

andrzejnovak:
@kratsg Changing the fonts that experiments chose is outside of the scope here. Considering that, if I interpret your comment correctly you are rather for the fonts failing loudly than silently. That's fine by me.

PR #174, opened 2020-08-13
fix: support latest hist label name

henryiii:
Hist renamed "title" to "label" following feedback. For now, support both.

### DIFF
--- a/mplhep/utils.py
+++ b/mplhep/utils.py
@@ -1,7 +1,10 @@
 def get_histogram_axes_title(axis):
     # type: (object) -> str
 
-    if hasattr(axis, "title"):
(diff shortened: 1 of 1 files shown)


andrzejnovak:
Will merge, just sorting out some tag creator problems

henryiii:
This is why I recommend a simple, clean system in https://scikit-hep.org/developer instead of lots of custom bots. :)

andrzejnovak:
@henryiii Sorry for the spam :) The setup should be working now, but it failed here since both branches needed to be up to date. Overall I think it's worth some initial pains.  The new changes should also allow it to work on PRs from outside contributors which was previously my biggest reservation.

PR #183, opened 2020-09-05
feat: include ALICE style 

fmazzasc:
Hello,
with this P.R. I would like to introduce a first draft of the ALICE style. I have also updated the README.md and the tests consequently: therefore a new figure in the ` baseline` repo is generated


mplhep/alice.py
```diff
@@ -0,0 +1,43 @@
...
+@deprecate.deprecate("Naming convention is changing. Use ``mplhep.alice.label``.")
+def alicelabel(*args, **kwargs):
+    return _alice_label(*args, **kwargs)
+
+
+@deprecate.deprecate("Naming convention is changing. Use ``mplhep.alice.text``.")
+def alicetext(**kwargs):
+    return _alice_text(**kwargs)
+
+
```
andrzejnovak, line 35: You can remove these, no need to add new stuff in an already deprecated mode.

andrzejnovak:
Welcome @fmazzasc, thanks for the PR! Happy the see we'll have the full set now :)

mplhep/alice.py
fmazzasc: Done :)

Issue #177, opened 2020-08-18
Bug in cms rlabel when using blank year


mondalspandan:
```hep.cms.label(ax=ax,paper=True,data=True,lumi=41.54,year="")```
leaves a comma between the lumisection and the CoM energy.

PR #184, opened 2020-09-10
fix: #177

andrzejnovak:


andrzejnovak:
@matthewfeickert any idea why papermill started to act out?

matthewfeickert:
> @matthewfeickert any idea why papermill started to act out?

No, given that it is complaining about something internal and there hasn't been a new release [since August 17th](https://pypi.org/project/papermill/2.1.3/#history). I restarted the jobs to see if it was a bug, but if it continues to happen then I would replicate locally, make a minimal failing example, and then open an Issue with them.

matthewfeickert:
Though seems that you're not the only one seeing it, as this was mentioned in a papermill issue https://github.com/nteract/papermill/issues/289 back in 2019 and is now being seen again by the OP of the issue at https://github.com/yaananth/run-notebook/issues/11.

PR #185, opened 2020-09-13
feat: clean up for minor version bump

andrzejnovak:


andrzejnovak:
@henryiii Could you take a look and tell me if this is along the lines of what you imagined for the return types? https://github.com/scikit-hep/mplhep/pull/185/commits/94279197e8f73718a9cf819b774e84911403654e

henryiii:
Yes, I think so. You could do two named tuples, and select between them based on error bars being present. Returning a variable length tuple is not ideal directly from the function, but since you are returning a list anyway, having two different types of tuples in the list might be better than having tuples with Nones (I'm not sure).

Any chance of getting Python 3 typing into the repo, actually (not necessarily right now)? Hist is typed; it's quite nice to be able to look at the signature and see that it returns `List[StepArtists]` (or `List[Union[StepArtist, StepArtists]]`).

andrzejnovak:
Is there anything problematic about tuples with `None`s in them? It feels more stable to work with than something where the return varies based on if errors are plotted. It's already a bit annoying the return will change based on "histtype", but I don't think there's a way around that. 

I've meant to bring it up to speed with the recommendations you have in skhep/developer including typing, but haven't had the time yet.

henryiii:
It's a toss up. Depends on which usage reads better.

PR #195, opened 2020-09-25
ci: fixes and packaging updates

henryiii:
Clean up some distribution related code.

- ci: use pre-commit in CI (already supported, just not used in ci)
- ci: allow SDist/wheels to be downloaded from any run
- ci: fix issue with broken publish job
- ci: use python-build instead of pep517.build
- refactor: move to src directory layout
- refactor: use declarative config file for most metadata
- refactor: use setup.cfg for pytest (saves a file in the main page)

Minor other cleanup.


.github/workflows/publish-package.yml
```diff
@@ -20,27 +20,34 @@ jobs:
...
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.7
-    - name: Install pep517
+
+    - name: Install build
       run: |
-        python -m pip install pep517 --user
+        python -m pip install build
```
henryiii, line 37: pep517 is not intended to be a user-facing tool. See https://github.com/pypa/pep517/issues/91

.github/workflows/ci.yml
```diff
@@ -3,75 +3,50 @@ name: CI/CD
...
 jobs:
   linter:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        python-version: [3.7]
-    name: linter - flake8 - python ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    name: pre-commit (linter)
```
henryiii, line 15: This runs black, flake8, and a few more checks,  provides nice error messages, and has custom caching built in, so it's lightning fast.

.github/workflows/publish-package.yml
```diff
@@ -20,27 +20,39 @@ jobs:
...
+
     - name: Build a binary wheel and a source tarball
       run: |
-        python -m pep517.build --source --binary --out-dir dist/ .
+        python -m build
+
+    - name: Make files available as an artifact
+      uses: actions/upload-artifact@v2
+      with:
+        path: dist/*
```
henryiii, line 46: This way you can download and check the SDist/wheel directly from Actions.

PR #196, opened 2020-09-26
style: put flake8 config in setup.cfg

henryiii:
The flake8 config is usually in setup.cfg. Saves a main folder file, and is a common place to put this.

You _can_ put the bump version config in setup.cfg too if you want to, but it would change some CI I expect, and it does often sit in its own file (partially because bump version changes this file when it runs).


henryiii:
What is `.restyled.yaml` used for?

PR #218, opened 2020-11-03
feat: Add styles module

matthewfeickert:
In an effort to keep the library directory a bit more clean, move all of the `style_X` code files into a `styles` module. As this is basically just changing import paths none of the public API should change.

```
* Change imports to using .styles
* Move code from styles.py to styles/__init__.py
```


matthewfeickert:
@andrzejnovak I didn't also move the different experiment named files under the `styles` module, but if you think they should then I can easily do that here as well.

andrzejnovak:
I would prefer those to stay as-is, so that one doesn't have to dig around modules to get the style/label they want.

PR #221, opened 2020-11-05
fix: 2D hist labels only assert the same shape between labels and hist

pfackeldey:
Hi,

This change allows to write anything into the bin as a label.
Consider the case:

```python
In [2]: H
Out[2]:
array([[0.082, 0.286, 0.397, 0.448, 0.472, 0.501, 0.528, 0.557, 0.592, 0.631],
       [0.074, 0.257, 0.38 , 0.447, 0.47 , 0.495, 0.521, 0.53 , 0.527, 0.536],
       [0.03 , 0.157, 0.262, 0.341, 0.356, 0.388, 0.421, 0.444, 0.464, 0.493],
       [0.023, 0.123, 0.214, 0.288, 0.318, 0.331, 0.354, 0.367, 0.419, 0.412]], dtype=float32)

In [4]: labels
Out[4]:
array([['$0.08\\pm0.00$', '$0.29\\pm0.00$', '$0.40\\pm0.00$', '$0.45\\pm0.00$', '$0.47\\pm0.01$', '$0.50\\pm0.00$', '$0.53\\pm0.00$', '$0.56\\pm0.00$', '$0.59\\pm0.00$', '$0.63\\pm0.01$'],
       ['$0.07\\pm0.00$', '$0.26\\pm0.00$', '$0.38\\pm0.00$', '$0.45\\pm0.00$', '$0.47\\pm0.00$', '$0.50\\pm0.00$', '$0.52\\pm0.00$', '$0.53\\pm0.00$', '$0.53\\pm0.00$', '$0.54\\pm0.01$'],
       ['$0.03\\pm0.00$', '$0.16\\pm0.00$', '$0.26\\pm0.00$', '$0.34\\pm0.02$', '$0.36\\pm0.00$', '$0.39\\pm0.00$', '$0.42\\pm0.00$', '$0.44\\pm0.00$', '$0.46\\pm0.00$', '$0.49\\pm0.00$'],
       ['$0.02\\pm0.00$', '$0.12\\pm0.00$', '$0.21\\pm0.00$', '$0.29\\pm0.01$', '$0.32\\pm0.01$', '$0.33\\pm0.01$', '$0.35\\pm0.00$', '$0.37\\pm0.01$', '$0.42\\pm0.02$', '$0.41\\pm0.01$']],
      dtype='<U13')

...
hep.hist2dplot(H, xaxis, yaxis, labels=labels)
```
Now this change would allow to use these labels as the text on each bin. The only requirement is that the shapes are equal. `np.array_equiv` also checks that the elements are the same, which does not allow this kind of labelling freedom right now.

I hope I didn't miss something here.

Best, Peter


andrzejnovak:
Thanks! Indeed this was how it was intended to work :) Could you add a test for it, while you're at it?

pfackeldey:
Sure, could you point me to the test (and maybe to the reordering)? I am unfamiliar with unittest of mpl plots.

andrzejnovak:
Something like this:
https://github.com/scikit-hep/mplhep/blob/master/tests/test_basic.py#L143

To run tests:
`py.test --mpl`

To generate a new set of baseline images
`py.test --mpl-generate-path=tests/baseline`

An image like this https://github.com/scikit-hep/mplhep/blob/master/tests/baseline/test_hist2dplot.png gets created and needs to be added to the PR so that the tests run.

pfackeldey:
alright, I hope this is ok like this :)

andrzejnovak:
Looks good, thanks.

PR #222, opened 2020-11-09
feat: add LHCb style 2

mayou36:
A new LHCb style that is used by many people but just floating around.

There is the LHCb and LHCbTex; however in the style, the techversion is meant to be set explicitly as the default. Further assuming that latex is basically everywhere installed, it seems reasonable to me to change the default from `LHCb` and `LHCbTex` to `LHCb` and `LHCbNoTex`. Or what are your thoughts on this? Should the non-tech version remain the default?

For the numbering of LHCb versions: Does it make sense to have an LHCb1 and LHCb2 style instead of LHCb and LHCb2? For example for the string setter. Maybe over time we can figure out which one is preferred.

Any other thoughts?

Main difference:
- Decreased linewidth
- automatic minor ticks on
- ticks on all axes
- unicode minus (although bug currently?)

Comparison: old/lhcb (top) vs new/lhcb2 (bottom)

lhcb1_default

myplot


tests/test_styles.py
```diff
@@ -1,9 +1,9 @@
 import os
 import sys
-import pytest
```
andrzejnovak, line 3: No reason to reorder this right?

src/mplhep/styles/lhcb2.py
```diff
@@ -0,0 +1,115 @@
...
+    "axes.labelsize": 32,
+    "axes.linewidth": 2,
+    "axes.facecolor": "white",
+    # Custom colors
+    "axes.prop_cycle": cycler("color", colors),
+    # + cycler("marker", markers)  # TODO: markers?
+    "axes.formatter.min_exponent": 3,
+    # Figure properties
+    "figure.figsize": (8, 8),
+    "figure.dpi": 2500,
```
andrzejnovak, line 47: This is a very strange choice of dpi, leading to a huge file.

andrzejnovak (commented):
- There seems to be an empty file added `src/mplhep/styles.py`. 
- Could you add the the PR description a comparison between the original and the new style
- I think for the structure, there shouldn't be a new file, but the new style dicts should be part of `src/mplhep/styles/lhcb.py`

andrzejnovak:
Regarding the name, if there are any substantial differences being introduced in this style maybe there's a good descriptive word for that?

I don't have a big stake in this, not being in LHCb, but TeX is certainly not installed everywhere, so I don't think is should be the default.  Also since the only item in the preamble is loading Times New Roman, is there actually any difference in the rendering? It's worth noting that `"usetex": True` makes rending plot noticeable slower.

src/mplhep/stylelib/LHCb2.mplstyle
```diff
@@ -0,0 +1,91 @@
+## LHCb-like style
```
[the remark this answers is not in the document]
mayou36: Removed

tests/test_styles.py
mayou36: There is: [PEP 8 suggests](https://www.python.org/dev/peps/pep-0008/#imports) an ordering of standard library, 3rd party library, own libraries/helper

src/mplhep/styles/lhcb2.py
mayou36: yep, true, I reduced that to 300

src/mplhep/lhcb.py
```diff
@@ -1,58 +1,9 @@
...
-    return label_base._exp_text(
-        "LHCb",
-        text=text,
-        italic=(False, False),
-        fontsize=28,
-        fontname="Times New Roman",
-        loc=1,
-        **kwargs
-    )
+from .lhcb_common import label, text
```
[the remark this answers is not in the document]
andrzejnovak: I think the API for the elements related to this PR should be
```
mplhep.style.LHCb
mplhep.style.LHCb2
mplhep.lhcb.style.LHCb
mplhep.lhcb.style.LHCb2
mplhep.lhcb.label()
mplhep.lhcb.text()
```

mayou36:
> * There seems to be an empty file added `src/mplhep/styles.py`.

removed

> * Could you add the the PR description a comparison between the original and the new style

doing

> * I think for the structure, there shouldn't be a new file, but the new style dicts should be part of `src/mplhep/styles/lhcb.py`

but what about the attribute `mplhep.<experiment>.style`? this points currently to the style module, but I think it is not needed actually. Remove it?

andrzejnovak:
> but what about the attribute mplhep.<experiment>.style? this points currently to the style module, but I think it is not needed actually. Remove it?

That's just and an additional way to call the dict. `<experiment>.style` points to the module, not to the specific dict. You could do sth like
```
import mplhep.lhcb as lhcb
plt.style.use(lhcb.style.LHCb)
# or
plt.style.use(lhcb.style.LHCb2)
lhcb.label(ax)
```
You could make an argument along the zen of python "There should be only one way of doing things" but that should not be part of this PR.

src/mplhep/lhcb.py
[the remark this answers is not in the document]
mayou36: yep, updated

andrzejnovak:
Cool, thanks for the updated description. The new style looks much better balanced to me. The problem with the minus sign is usually caused by the unicode minus not being part of the font character set, but it should be a part of Times New Roman. Do you get any fallback warnings when you make the plot?

You can tell matplotlib to use the dash instead `"axes.unicode_minus": False`

mayou36:
Yes, agree. And nope, I don't get a fallback warning, just `RuntimeWarning: Glyph 8722 missing from current font.`

I'm trying to investigate. It can be swapped of (as in LHCb1 style), however the minus is then way shorter.

andrzejnovak:
That's quite strange since it should be a part of Times New Roman and also mpl shouldn't silently change to a different font. Good luck investigating.

mayou36:
Yep, I've opened an [issue](https://github.com/matplotlib/matplotlib/issues/18926), even rcParams don't seem to work properly... somethings somewhere fishy

andrzejnovak:
Are you sure you have both times (via msftcorefonts) and tahoma (don't know where the ships) installed and available? You can double check with the mpl font manager and manually register them like here https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/__init__.py#L67

mayou36:
> Are you sure you have both times (via msftcorefonts) and tahoma (don't know where the ships) installed and available? You can double check with the mpl font manager and manually register them like here [https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/__init__.py#L67](https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/__init__.py?rgh-link-date=2020-11-10T16%3A54%3A25Z#L67)

... you are right, I simply did not install it. Didn't thought I don't have "Times New Roman"

Then everything works for me

src/mplhep/styles/__init__.py
```diff
@@ -1,11 +1,13 @@
 import sys
+
 from matplotlib.pyplot import style as plt_style
 
+from .alice import ALICE
+from .atlas import ATLAS
+
 # Short cut to all styles
```
andrzejnovak, line 8: this is a bit confusing, they can be ordered alphabetically, but why is the comment left stranded in the middle

andrzejnovak (commented):
Ok, looks good. One minor comment left and for some reason this PR removes `tests/baseline/test_hist2dplot_custom_labels.png` which surprisingly does not crash the tests, can you undo the change?

andrzejnovak:
> Only thing: I still think it could be useful to use LHCb1 and LHCb2 styles and omit the `LHCb` one? Over time we may get LHCb3 as well (or `LHCb` using the "best" available).
> Or what is in general the idea of improving styles?

I think we want the one that's closest to official the default i.e `LHCb`. I think the way towards that without causing much grief to anyone already using it is indeed to have `LHCb1` and `LHCb2` and alias `LHCb1` for now as `LHCb` with a message that the style will change in the future since the LHCb2 looks to me more correct (but! I am not part of LHCb). You could add that in this PR still if you like.

src/mplhep/styles/__init__.py
mayou36: Ah sure, that was my autoformatter in PyCharm. I've moved the comment up.

mayou36:
> Ok, looks good. One minor comment left and for some reason this PR removes `tests/baseline/test_hist2dplot_custom_labels.png` which surprisingly does not crash the tests, can you undo the change?

It didn't get removed, just changed and removed 32 bits from the file. Don't know what or why, but reverted it.

mayou36:
> I think we want the one that's closest to official the default i.e `LHCb`. I think the way towards that without causing much grief to anyone already using it is indeed to have `LHCb1` and `LHCb2` and alias `LHCb1` for now as `LHCb` with a message that the style will change in the future since the LHCb2 looks to me more correct (but! I am not part of LHCb). You could add that in this PR still if you like.

Fully agree, that was exactly also my idea. LHCb1 and LHCb2 will be guaranteed to not change, while LHCb may does in order to get the best style.

Yes good, I'm gonna add it in this PR, let you know once done.

mayou36:
I've changed the README to use `LHCb2` in the example with an `  # or "LHCb1"` in order to advertise the new style yet also hinting at the other possibility.

Just approve if you're fine with it

PR #223, opened 2020-11-11
refactor: deprecation dict

mayou36:
In order for the LHCb style to be properly deprecated, we need to have the `deprecation_dict` working correctly.

The current implementation seemed to have minor non-optimalities though that are fixed with this PR:
- it is not a subclass of `dict`, meaning it doesn't react on the `isinstance(..., dict)`. Of course, the golden way is to check not the class but the interface. However, we do it ourselves in the test (that's why the test currently fails in #222) and I would not expect the average user to do it correctly. So we have to assume that `isinstance(depr_dict, dict)` has to work.
- The filter was mangled in order to print deprecation warnings. While it is somehow a pity that we cannot make the user see it, setting global filters and _resetting_ them to `'default'` should not be the way to go. E.g. if a user has set the filter to something different, we change it not to 'default'. I would propose to have a variable Warning as an argument to the `deprecation_dict` (e.g. FutureWarning) that isn't suppressed, so we have the option to use non-suppressed warnings.
- the warning would be on _every time_ an element is accessed - however, often we want to print it just once.
- `key transform` was removed. Was there a specific reason for this? Furthermore, new dunder methods (`__keytransform__` is not a method that exists in Python, correct?) should not be introduced by the user as they are reserved for possible future Python usage.

The things above have been implemented.


andrzejnovak:
Cool, I'll merge it when  you fix black in the precommit

mayou36:
Done ;)

mayou36:
Done ;)

However, we have a changed behavior now as the `DeprecationWarnings` won't show anymore... however, this is actually more of a feature of the language than an issue on our side.

andrzejnovak:
One last thing. Is it possible to get the "current" state of the deprec warning and return to that after emitting it here with always, instead of forcing the default?

mayou36:
While that may be possible, I think you want to keep the warning showing up, right? Mangling with the filter is though dangerous: if a user wants to suppress some warnings now (e.g. _explicitly_ DeprecationWarnings) he can _not_, so we should not do it.
I'd propose to rather go for a [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning) but also show it only once. Should I just change it accordingly that it always shows the warning once and emits a `FutureWarning`?

andrzejnovak:
Yeah, i think that's a good idea.

mayou36:
Thinking again, I would rather not do it, but use `FutureWarning` for when something changes, such as the LHCb style that may changes (it is not forbidden to use LHCb, it will just maybe change in the future) compared to a deprecation like `ROOT` which will be removed.

The advantage is that DeprecationWarnings can be detected by smart IDEs and by unittests. If the user does not see them, well... they can still notice once the feature is removed and an error is raised with an instruction on how to update.

So: FutureWarning is maybe seen more (just talking about e.g. ROOT deprecation, for LHCb we anyway should use FutureWarning), but is not the correct one and automatic tools won't detect it. I tend to DeprecationWarnings, but no strong opinion. What's your call?

andrzejnovak:
I think to first order, there should be a visible warning. Deprec is a more appropriate word, but if changing the visibility level is bad practice, then we should just switch to FutureWarning

> FutureWarning: always displayed by default, recommended for warnings intended to be seen by application end users (e.g. for deprecated application configuration settings).

I think this is our use case.

andrzejnovak:
@mayou36 is this ready to be merged?

mayou36:
Now yes. Otherwise you can also just approve it and I'll merge when things have passed

Issue #225, opened 2020-11-11
Hardcoded error defaults shadows rc settings


mayou36:
The defaults of an error plot are hardcoded inside the function:
https://github.com/scikit-hep/mplhep/blob/286639cf9bcf98cf26e8be55461ce616649ca8c7/src/mplhep/plot.py#L382

This renders style setting useless for this plot. Can we somehow pass them through, define them as well in the style settings? Or just not use the defaults? Currently, a style cannot change this.

andrzejnovak:
Some defaults are needed, otherwise a plot that should look like this without any applied style
![image](https://user-images.githubusercontent.com/13226500/98830926-3ec99200-243b-11eb-99dc-ad2e08694b6b.png)
ends up looking like this
![image](https://user-images.githubusercontent.com/13226500/98830893-35402a00-243b-11eb-9ea1-c88864ff2649.png)

We could probably extend the checking to see if any rcParams are modified from the mpl baseline and if so, don't overwrite them

mayou36:
Yes, I fully agree on the defaults to be there. Maybe we can also just add "other fields" to the rc parameters? Basically defaults for mplhep and not for matplotlib which are though in the same style dict?

But that's maybe a second order problem, I agree to go with checking the style dict first.

What is problematic about this: that the markers etc are often with a lineplot in mind, not with an errorbar plot. So these seem two different things actually.

andrzejnovak:
> Yes, I fully agree on the defaults to be there. Maybe we can also just add "other fields" to the rc parameters? Basically defaults for mplhep and not for matplotlib which are though in the same style dict?

This would alter the API, because you could no longer pass the dict directly to `plt.style.use`


> What is problematic about this: that the markers etc are often with a lineplot in mind, not with an errorbar plot. So these seem two different things actually.

Are any of the current defaults in conflict because of that? I don't think so and anyway this would have to be solved in mpl.

andrzejnovak:
This is fixed https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/plot.py#L401

PR #226, opened 2020-11-11
fix: add LHCb2 errorbar plot style finalized

mayou36:
I've added a small update (sorry, that glitched before) which was somewhat hidden due to a possible "bug" (or feature?) in histplot. This settings would give nice errorbar plots (in LHCb-style).

The comparison only works by explicitly giving the params as kwargs, not all are used from the style, see #225 

Comparison: old (top) vs new (bottom)

image

image


andrzejnovak:
Yeah, the default params when plotting as `'errorbar` should probably first check the rcParams https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/plot.py#L382

PR #236, opened 2020-12-10
docs: Add the new LLP->emunu paper to the list of analysis using mplhep

marinang:
This adds the following paper https://arxiv.org/abs/2012.02696 to the list *Use in publications*.

I want also would like to thank the `mplhep` authors/contributors for designing this package that helped me designing nice plots with the LHCb style.


andrzejnovak:
Awesome, thanks a lot for including the reference!

PR #239, opened 2021-01-08
fix: ATLAS sizing and styles

kratsg:
This also adds two new styles. So now `ATLAS` has three styles:

* `ATLAS`: this will rely on math fontset `TeX Gyre Heros` (looks nicer overall, but sucks for `\chi`)
* `ATLASAlt`: this will rely on dejavusans (default math fontset)
* `ATLASTex`: this will rely on LaTeX (`usetex` is enabled)

I've done a lot of cross-checks and things look quite a bit more consistent now and hard to tell apart from ROOT-related versions.


matthewfeickert:
@kratsg Can you also post some images here to show the results of this PR? Maybe also some comparisons between ROOT and this PR's effect?

kratsg:
> @kratsg Can you also post some images here to show the results of this PR? Maybe also some comparisons between ROOT and this PR's effect?

I can't. They're internal for me. The tests should have baseline comparisons, but I cannot get the tests working on my machine anyway...

matthewfeickert:
> I can't. They're internal for me. The tests should have baseline comparisons, but I cannot get the tests working on my machine anyway...

Ah right, wasn't thinking there. I'll see if I can get the tests going on Ubuntu.

andrzejnovak:
NIce. Thanks a lot @matthewfeickert, I was only gonna get to this on the weekend. @kratsg tests look ok now, is this ready to merge?

matthewfeickert:
> NIce. Thanks a lot @matthewfeickert, I was only gonna get to this on the weekend.

Hold up a sec, as I realized that I updated _all_ of the test baselines when I might not have needed to. I force-pushed with only the ATLAS one updated, so let's see if that's good now.

matthewfeickert (approved):
Okay tests are all passing and we only update the `tests/baseline/test_style_atlas.png` test file instead of all of them now. :+1: I'm good to go on this, so thanks @kratsg for making this! More beautiful ATLAS plots in the future. :)

andrzejnovak:
In principle, if the baselines are 0 bytes diff it should be ok. I am not actually very clear on why they register in git sometimes

matthewfeickert:
> In principle, if the baselines are 0 bytes diff it should be ok. I am not actually very clear on why they register in git sometimes

Sure. Though in general I find these diffs confusing to people in the future, so I like to avoid them when possible. :)

PR #251, opened 2021-01-29
Feat/protocol

### LINKED ISSUE #246: [feat] Implement PlottableProtocol
- https://github.com/scikit-hep/boost-histogram/pull/476
- https://github.com/henryiii/uhi/pull/2

henryiii:
Adding PlottableHistogram. Refactoring code a bit.

Closes #246.

Smaller changes:
* Added 3.9 to the test matrix, dropped a couple of in-between macOS test runners to compensate
* Show all reasonable things in the PyTest summary, not just xfails. Fails are more interesting than xfails!
* Use MyPy, and make it pass - should be expanded later with more strictness options and more types!
    * `__all__` must be strings
    * Name error try/except was useless, because the next line with open does not allow None
    * Should use `np.sum` instead of `sum` for NumPy
    * Added just enough explicit types to get MyPy to pass
* Change py.test -> pytest, the old name should not be used anymore (more-or-less deprecated)
* Add Scikit-HEP as a maintainer.
* Added pytest-mock, with some basic mock tests - maybe expand eventually? Can track exactly what gets passed to mpl.
* Test to ensure LHCb does raise a warning, and avoid warning showing up in the tests
* Avoided having a test named the same thing as the file it's in, makes it hard to run by itself with `-k <name>`.
* Dropped a few asserts. Assert should be used to check something for correctness only; it can be disabled in production. Assert is also great for telling mypy that a type is known.
* Fixed a few spelling errors that VI pointed out. Probably made more.

Then, the plot/utils files were heavily refactored. There was _one_ test change: `histplot(uproot)` now plots error bars, since uproot does not return None for `.variance()`. I've explicitly passed `yerr=False` to disable the error bar plotting (a good test to have, anyway); one with error bars would likely be good to add.

Key changes:

* Refactored plotting to make it type safe and avoid mega-functions. Now, Lists of hists get handled uniformly, with a single hist simply being passed as `[h]` to unified plotting code, and I broke up functions to some extent.
* Added a _few_ types, more should be added, _especially_ in public interfaces. Once interfaces are typed, we can add a `py.typed` token file and this will become externally statically typed!
* At type check time, uhi, NumPy 1.20+, and mypy 0.800+ are required (ensured by pre-commit)
* I've used string types to avoid run time dependencies, and to keep compatibility with Python 3.6. The strings can go away when we go to 3.7+ and add `from __future__ import typing`.
* I've pulled out functions-in-functions. They are hard to read, as you have to scan the whole thing looking for variable capture. Much better to make it explicit.

Follow on ideas:

* The files need to be broken up a bit, there's too much in each file.
* The plothist function can still have several portions pulled out into helpers. The more pieces, the easier testing is, the easier it is to reason about what it does (as long as they are clear, well defined concepts).
* One really nice feature from Pandas is `plot.<stuff>`. We could have `plot.fill`, `plot.step`, and `plot.errorbar`, each with tighter API and more focused docs than the general function.
* I've converted to `<stuff>` -> `List[PlottableProtocol]` objects (with a quick, NumPy implementation that satisfies the Protocol). However, I then go back to the old NumPy arrays so that I could use the old code. Ideally, we should only use the Protocol object after that, would be cleaner.
* I left off handing discrete bins. That should be added in eventually, with thought about labels and such (that's the source of most of the `type: ignore`'s, MyPy forces you to implement the whole Protocol otherwise).
* Explicit support for `Kind = "MEAN"` would be nice, currently this is a "dumb" plotter as we called it, treating both kinds of histograms the same.
* We really need more tests for the new stuff (PlottableProtocol's). But we don't have a released boost-histogram with the protocol yet, only in master. Something from Uproot4 should be good? Uproot4 should conform to the Protocol already (eventually I might be able to get @jpivarski to add typing and have it checked explicitly, who knows :) Will be post Python 2, though).

@andrzejnovak , ready for review! Let me know if I need to resquash my history or anything.


henryiii:
@andrzejnovak is there a way to see what's plotted and compared in the diff? (assuming I"ll need to start up a linux image to check it, as it's disabled for macOS I think? :/ ) Plots look fine locally (save for having text on them, which I'm guessing is removed from the solutions to avoid font variations?), and I've added a mocking unit test which seems to indicate the same matplotlib calls are being preformed (only added the simple case so far).

andrzejnovak:
@henryiii I am not sure I understood right, but running the test suite with `py.test --mpl` should produce the figures where marked in the test. This nominally produces the ref/new and diff in a temp folder or you can point it to a specific one.

The only limit in the current test setup are font related tests, because I don't know how to install the required fonts to the macOS image in GHA

andrzejnovak:
Actually I should look into https://docs.github.com/en/actions/guides/building-and-testing-python#packaging-workflow-data-as-artifacts. It would be really cool to see the diff images online.

henryiii:
Yes, that's what I was asking, thanks! What folder is the output in? I could just add that instead of running locally.

henryiii:
Ahh, I probably renamed the tests (since matching the file name and the test name makes it really hard to run just one test with `-k`), and didn't realize that the mpl plug-in was matching on test name. So I should have mocked one of the other tests.

andrzejnovak:
By default the results go to `/tmp` but you can change it with `  --mpl-results-path=MPL_RESULTS_PATH` If you are adding new tests you want to run 
```
py.test --mpl-generate-path=tests/baseline
```
One annoying thing is that this often registers as changed binary files even if the test would pass as no diff. So you only want to add the new baselines to the commit.

setup.cfg
```diff
@@ -16,6 +18,7 @@ classifiers =
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
```
andrzejnovak, line 21: Should be added to test matrix

tests/test_styles.py
```diff
@@ -11,7 +11,7 @@
 
 """
 To test run:
-py.test --mpl
+b
```
andrzejnovak, line 14: typo

henryiii: That's weird, most of my typos like this are `:w` :) Thanks!

setup.cfg
henryiii: Yes, good point, will add after I get the current code passing tests :)

tests/test_styles.py
henryiii: By the way, "pytest" is the recommended entry point for PyTest since version 3 (though py.test may remain around forever). The old `py.test` naming was from an era when they planned to have several things in a `py` namespace; `py.path` has been mostly replaced by pathlib and is no longer recommended standalone, and the focus is on pytest, so they dropped py.test from the docs and entry points by version 3.

henryiii:
New artifacts for the test work great, by the way! Very helpful :)

andrzejnovak:
This is great! Thanks a lot! 

One thing I was wondering looking at the changes it might make it cleaner still if external errors (especially sumw2) got parsed into the internal `[h]` object list and picked from it when the "display" errors are calculated, but it certainly doesn't need to come with this PR.

Regarding API of mplhep.{hist|errorbar|fill|step} I am not sure how much sense it makes to multiple functions with essentially the same signature. We also can't use it to simplify the return artists all that much since hist+errors will still be two objects with current mpl. But we can take this to Discussions if you like.

PR #254, opened 2021-02-05
ci: upload py.test --mpl artifacts

andrzejnovak:


andrzejnovak:
cc @henryiii you should be able to download pytest artifacts from GHA after this

henryiii:
Perfect, that's what I was thinking. :)

PR #255, opened 2021-02-09
ci: drop pre-commit job

henryiii:
Failed due to cache issue - we already use pre-commit.ci here, so let's just use that, it's faster and can auto-correct in some cases.


henryiii:
This is a ci change. It should not be a release, no code at all has changed.

PR #256, opened 2021-02-11
feat:Different scale factors to scale the y axis for linear and log y scales to remove legend overlap

bsathian:
1.05 is too small a number if the y axis is set to log. To get the equivalent effect, the y axis needs to be scaled up by 10**(1.05). I check if the y axis is log and rewrite the scale number


bsathian:
I'm closing this so that I can follow the instructions provided in CONTRIBUTING.md and open a pull request that adheres to the guidelines

PR #258, opened 2021-02-11
fix: adjust .yscale() for log axis

bsathian:
1.05 is too small a number if the y axis is set to log. To get the equivalent effect, the y axis needs to be scaled up by 10**(1.05). I check if the y axis is log and rewrite the scale number


andrzejnovak:
Thanks a lot @bsathian.

Issue #241, opened 2021-01-11
Add custom Font Manager to download fonts


matthewfeickert:
Following [discussion in Issue #240](https://github.com/scikit-hep/mplhep/issues/240#issuecomment-758098060), it seems there is agreement that adding a [custom Font Manager similar to how Colin Carroll has done in `ridge-map`](https://github.com/ColCarroll/ridge_map/blob/e7331d5038e3dee99aa061d562a8661cac6b5d62/ridge_map/ridge_map.py#L16-L53) (thanks @ColCarroll!) would be a useful thing for `mplhep` and sidesteps most of the need to vendor fonts with the library.

It should be relatively straightforward to adopt what Colin setup in `ridge-map` for `mplhep`'s needs.

(For posterity: [The Tweet that directed us to this information](https://twitter.com/colindcarroll/status/1348434308701315073?s=20))

andrzejnovak:
Close via #282 - fonts in a separate package, util out of scope now

PR #263, opened 2021-02-14
feat: get fonts by download, don't package them

andrzejnovak:
@matthewfeickert first take on #241, let me know if you think there should be a better way to fetch the fonts than one by one. Unfortunately, there doesn't seem to be a unified way to fetch them from their original repositories, so we'll still have to host copies, but the pypi dists should be quite a bit lighter.

Actually, on second thought, this could lead to potential problems, if one for created an image/env on while having internet connection, but only tried to load it later when not connected to the internet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Package specific
+src/mplhep/fonts/downloaded/*
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,3 @@
 include src/mplhep/.VERSION
-include src/mplhep/stylelib/*.mplstyle
-recursive-include src/mplhep/fonts/ *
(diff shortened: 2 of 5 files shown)


andrzejnovak:
Ok, after letting it sit for a while, the fonts should probably go into a separate package like `mplhep_data`

matthewfeickert:
> Actually, on second thought, this could lead to potential problems, if one for created an image/env on while having internet connection, but only tried to load it later when not connected to the internet.

Sorry for being slow here. I'll defer to you here. I think that the font issue is tricky in general, but IMO if you're making a container you're going to test it before you deploy it somewhere offline and if you see that it is downloading information when you **don't think it should** then the container owner is responsible for figuring out how to fix that. 

That being said, I think it is totally reasonable for you to not want to touch having to be responsible for fonts at all. But I think these helper functions are nice to have, but if kept maybe just made clear they aren't supposed to be relied on super heavily.

PR #277, opened 2021-03-16
refactor: use uhi's adaptor for PlottableHistogram

henryiii:
Building on https://github.com/scikit-hep/uhi/pull/13. Using slightly more "normal" Python 3.6+ typing.


henryiii:
I recommend using extras for docs, not requirements.txt. It's simpler and easier to maintain, and easier to test locally, and easier to ensure the module is always installed when the docs are running.

andrzejnovak:
Sorry, I thought I've already merged this in :/

PR #282, opened 2021-04-04
refactor!: separate out font files into mplhep_data

andrzejnovak:
@matthewfeickert with this PR we factor out fonts into separate package, which should 
a) downsize mplhep considerably so pypi hosting should be safe
b) allow us to vendor other fonts if needed/requested


matthewfeickert:
Sounds good

PR #292, opened 2021-05-04
ci: better message on autoupdate

henryiii:
See #268; this is now configurable!


andrzejnovak:
Awesome, thanks a lot!

PR #296, opened 2021-05-13
feat: savelabels()

andrzejnovak:
- Introduces `hep.savelabels()` to save copies with different labels. `hep.savelabels('test.png')` will produces 4 variation on experiment label:
    - "" -> "test_paper.png"
    - "Preliminary" -> "test_pas.png"
    - "Supplementary" -> "test_supp.png"
    - "Work in Progress" -> "test_wip.png"
- Allows `set_style()` to reset to default params
- Fixes italics for CMS label
- Modifies default ATLAS label to `loc=4` @kratsg @matthewfeickert this look right?
image


src/mplhep/label.py
```diff
@@ -346,3 +351,74 @@ def exp_label(
...
+    Example:
+        >>> import mplhep as hep
+        >>> hep.cms.label(data=False)
+        >>> hep.savelabels('test.png')
+
+        This will produces 4 variation on experiment label:
+            - "" -> "test_paper.png"
+            - "Preliminary" -> "test_pas.png"
+            - "Supplementary" -> "test_supp.png"
+            - "Work in Progress" -> "test_wip.png"
```
matthewfeickert, line 376: @andrzejnovak It would also be helpful if there was an "Internal" option.

matthewfeickert (changes requested):
One comment/request, but overall looks good.

src/mplhep/label.py
andrzejnovak: This is eniterly manually configurable by passing `labels=[...], suffixes=[...]`. On second thought 4 might be too many for a default, maybe nothing and "Preliminary" are more sensible. 

One thing I am not sure about is whether the API would be better to pass a dict like `{'label-text':'suffix/filename', ...}` instead of two arrays.

kratsg:
Just saw this, but the text size is wrong in the screenshot you have. Is that supposed to have gotten fixed?

andrzejnovak:
I see, the suffix and the "ATLAS" are supposed to be same fontsize?

kratsg:
Yes, also the spacing is too close.

PR #323, opened 2021-07-27
fix: update hist2dplot labels handling

### LINKED ISSUE #321: Inconsistent handling of `labels` argument for `hist2dplot`
# Issue 
The `labels` argument for [hist2dplot](https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/plot.py#L421) is expected to be treated as a `bool` type or histogram (`H`) type, according to it's documentation.  However, it only works as `bool` if it is `True`.

This is likely a result of mis-handling of `None` or handling the type of the user-provided `labels` argument.

# Expected behavior
Providing `labels = False` would lead to the default behavior of putting no labels on any of the bins of the 2D plot.

# Observed behavior

If you provide `labels = False`, as in:

```python
import hist
import numpy as np
import mplhep as hep

h2 = hist.Hist(
    hist.axis.Regular(10, 0.0, 1.0, label='Some label'),
    hist.axis.Regular(10, 0, 1)
)
h2.fill(np.random.normal(0.5, 0.2, 1000), np.random.normal(0.5, 0.2, 1000))
hep.hist2dplot(h2, labels=False, cbar=False)
```
you get the following error:
```
Traceback (most recent call last):
  File "foo.py", line bar, in <module>
    hep.hist2dplot(h2, labels=False, cbar=False)
  File "/foo/bar/env/lib/python3.8/site-packages/mplhep/plot.py", line 548, in hist2dplot
    elif H.shape == labels.T.shape:
AttributeError: 'bool' object has no attribute 'T'
```

dantrim:
Resolves #321

### DIFF
--- a/src/mplhep/plot.py
+++ b/src/mplhep/plot.py
@@ -542,16 +542,15 @@ def hist2dplot(
     plt.sca(ax)
 
     _labels: np.ndarray | None = None
-    if (type(labels) == bool) or (np.iterable(labels)):
-        if labels is True:
-            _labels = H
-        elif labels is False:
-            _labels = None
-        elif H.shape == np.asarray(labels).T.shape:
-            _labels = np.asarray(labels).T
+    if isinstance(labels, bool):
+        _labels = H if labels else None
+    elif np.iterable(labels):
+        label_array = np.asarray(labels).T
+        if H.shape == label_array.shape:
+            _labels = label_array
         else:
             raise ValueError(
-                f"Labels input has incorrect shape  (expect: {H.shape}, got: {np.asarray(labels).T.shape})"
+                f"Labels input has incorrect shape (expect: {H.shape}, got: {label_array.shape})"
             )
     elif labels is not None:
         raise ValueError(
--- a/src/mplhep/plot.py
+++ b/src/mplhep/plot.py
@@ -541,9 +541,9 @@ def hist2dplot(
 
     plt.sca(ax)
 
-    _labels: np.ndarray
+    _labels: np.ndarray | None = None
     if (type(labels) == bool and labels is True) or (
-        isinstance(labels, (tuple, list, np.ndarray))
+        isinstance(labels, (list, np.ndarray))
     ):
         if labels is True:
             _labels = H
@@ -555,7 +555,7 @@ def hist2dplot(
             )
 
     text_artists = []
-    if len(_labels) > 0:
+    if _labels is not None:
         for ix, xc in enumerate(xbin_centers):
             for iy, yc in enumerate(ybin_centers):
                 color = (
--- a/src/mplhep/plot.py
+++ b/src/mplhep/plot.py
@@ -541,17 +541,21 @@ def hist2dplot(
 
     plt.sca(ax)
 
-    text_artists = []
-    if labels is not None:
+    _labels: np.ndarray
+    if (type(labels) == bool and labels is True) or (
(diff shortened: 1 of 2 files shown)


matthewfeickert (commented):
@dantrim Thanks for this PR. I think that everything is looking good here, but what would make this even better is if you could include a test. I think that the minimal failing example that you gave in Issue #321 would do nicely. :+1:

src/mplhep/plot.py
```diff
@@ -541,17 +541,25 @@ def hist2dplot(
 
     plt.sca(ax)
 
-    text_artists = []
-    if labels is not None:
+    _labels: np.ndarray | None = None
+    if (type(labels) == bool) or (np.iterable(labels)):
```
matthewfeickert, line 545:
```suggestion
    if isinstance(labels, bool) or (np.iterable(labels)):
```

I remember @henryiii telling me to always use `isinstance` instead of `type` to make sure things are checked correctly.

src/mplhep/plot.py
```diff
@@ -541,17 +541,25 @@ def hist2dplot(
...
-        elif labels is not False:
+        elif labels is False:
+            _labels = None
+        elif H.shape == np.asarray(labels).T.shape:
+            _labels = np.asarray(labels).T
+        else:
             raise ValueError(
-                "Labels not understood, either specify a bool or a Hist-like array"
+                f"Labels input has incorrect shape  (expect: {H.shape}, got: {np.asarray(labels).T.shape})"
             )
```
matthewfeickert, line 555: Could maybe simplify this logic a bit more with something like

```python
    _labels: np.ndarray | None = None
    if isinstance(labels, bool):
        _labels = H if labels else None
    elif np.iterable(labels):
        label_array = np.asarray(labels).T
        if H.shape == label_array.shape:
            _labels = label_array
        else:
            raise ValueError(
                f"Labels input has incorrect shape (expect: {H.shape}, got: {label_array.shape})"
            )
```

matthewfeickert (approved):
This all LGTM. @andrzejnovak This is ready for review from you now I think.

matthewfeickert:
So this PR will allow for something like the following

```python
# example.py
import hist
import numpy as np

import mplhep as hep

if __name__ == "__main__":
    np.random.seed(0)
    h1 = hist.Hist(
        hist.axis.Regular(10, 0.0, 1.0, label=r"$x$-axis label"),
        hist.axis.Regular(10, 0, 1, label=r"$y$-axis label"),
    )
    h1.fill(np.random.normal(0.5, 0.2, 1000), np.random.normal(0.5, 0.2, 1000))
    artists = hep.hist2dplot(h1, labels=False, cbar=False)

    ax = artists.pcolormesh.axes
    fig = ax.get_figure()
    fig.savefig("example.png")
```

giving

example

instead of the current error of

```pytb
Traceback (most recent call last):
  File "example.py", line 13, in <module>
    artists = hep.hist2dplot(h1, labels=False, cbar=False)
  File "/home/feickert/Code/GitHub/Scikit-HEP/mplhep/src/mplhep/plot.py", line 548, in hist2dplot
    elif H.shape == labels.T.shape:
AttributeError: 'bool' object has no attribute 'T'
```

andrzejnovak:
Great, thanks a lot for the fix @dantrim.

src/mplhep/plot.py
dantrim: Looks good :)

PR #328, opened 2021-08-05
feat: Allow xlabel/ylabel to be passed through with ax kwarg

### LINKED ISSUE #327: Allow passthrough of label through ax kwarg
This was discussed a bit on Slack with @andrzejnovak and @alexander-held, but given a comment by Thomas I thought I'd give an example as an Issue and see if the suggestion is viable in @andrzejnovak's mind.

## Problem:

```python
import numpy as np
import mplhep

import hist
from hist import Hist

np.random.seed(0)

hist_1 = Hist(hist.axis.Regular(10, 0.0, 1.0, name="x"))
hist_1.fill(np.random.random(1000))

fig, ax = plt.subplots()
mplhep.histplot(hist_1, ax=ax)
ax.set_xlabel("my label")  # want this to show up

hist_2 = Hist(hist.axis.Regular(10, 0.0, 1.0, name="x"))
hist_2.fill(np.random.random(1000))

# This overwrite the pervious label with the value of hist_2.axes[0].label, which is "x"
mplhep.histplot(hist_2, ax=ax)  # apparently the ax label info isn't used here?
fig.savefig("example.png", facecolor="white")
```

example

Thomas mentioned

> behavior of checking if `ax.get_label() != '`:  before being helpful and setting the label seems reasonable to me. This is more or less what Matplotlib does with auto limits (if the user call `set_xlim` or `set_ylim` we track that and stop trying to help, but if not, then we re-compute the autolimits every time you add an artist

So if this is agreeable what would be needed (and what I have on a branch) is changing

https://github.com/scikit-hep/mplhep/blob/055a9fb4769beb1c1a571d71e346c84ce23caa3e/src/mplhep/plot.py#L167

to
```diff
-    x_axes_label = get_histogram_axes_title(hists[0].axes[0])
+    x_axes_label = (
+        get_histogram_axes_title(hists[0].axes[0])
+        if ax.get_xlabel() == ""
+        else ax.get_xlabel()
+    )
```

With this in the example code gives the following instead

patch_example

@andrzejnovak  thoughts?

matthewfeickert:
Resolves #327

**Suggested squash and merge message**:
```
* Allow for existing axis xlabel and ylabel to be used to set axis labels if kwarg ax passed.
* Update mock count in tests to account for additional calls
```

### DIFF
--- a/src/mplhep/plot.py
+++ b/src/mplhep/plot.py
@@ -488,9 +488,6 @@ def hist2dplot(
         if not isinstance(ax, plt.Axes):
             raise ValueError("ax must be a matplotlib Axes object")
 
-    x_axes_label = ""
-    y_axes_label = ""
-
     hist = hist_object_handler(H, xbins, ybins)
 
     # TODO: use Histogram everywhere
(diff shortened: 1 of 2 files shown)


matthewfeickert:
Thanks @andrzejnovak. I know that you're busy now, but do you have thoughts in the future about how to implement the warning system given https://github.com/scikit-hep/mplhep/issues/327#issuecomment-894866868? If so, I'll open up a separate Issue from that comment and work on it.

andrzejnovak:
Sorry for not following up on that. I think as you currently implemented it the warning isn't necessary as the behaviour got changed to be non-destructive. 👍

PR #330, opened 2021-08-16
feat(LHCb)!: Use Tex Gyre Termes instead of Times New Roman

### LINKED ISSUE #273: LHCb2 style problems
I see two problems with the LHCb2 style : 
- Some minus signs don't show properly.
- 2 different fonts are used for the text in italics and the text not in italics. Is it possible to only use the one from the text not in italics ?
@mayou36 : Can you take a look please ?

jonas-eschle:
closes #273 

Switching to Tex Gyre Termes (can also be added in https://github.com/scikit-hep/mplhep_data/issues/6), which is free. Also using it for the math to be consistent.

This should resolve both issues raised in #273 

Also reduced the default DPI to be aligned with the Matplotlib default and the other styles (CMS,...). 300 DPI was too large.

I think the changes are fairly small and should not break code, LHCb3 would be too confusing for this minor changes (DPI is anyway manually adjusted if needed, the fonts are considered "equal enough").


jonas-eschle:
Ready from my side

andrzejnovak:
@jonas-eschle Cheers, thanks a lot for this. Sorry about the delay, I was on holidays

jonas-eschle:
No worries! Thanks for taking care of the library

PR #352, opened 2022-01-06
refactor: reorgnize histplot logic

andrzejnovak:


andrzejnovak:
@henryiii any idea why the mock test breaks? it looks like it's trying to "match" to the wrong call from the diff.

henryiii:
You FP'd it, is it working now? Or is there something I still should check?

andrzejnovak:
> You FP'd it, is it working now? Or is there something I still should check?

Got it in the end thanks, not very happy with mock error messages tho

PR #361, opened 2022-02-15
feat: Override CMS experiment name

pfs:
Enabling the option of overriding the experiment name, otherwise the code will fail with a duplicated argument error.
This is useful e.g. in case one would like to write CMS+Totem instead of CMS alone


andrzejnovak:
👍 thanks

PR #364, opened 2022-03-01
chore: remove "wheel" from build deps

henryiii:
Not needed via PEP 517, provided by `get_requires_for_build_wheel`.


andrzejnovak:
Thanks @henryiii

PR #365, opened 2022-03-04
feat: add center of mass energy kwarg

DickyChant:
Hi authors, 

When I use `mplhep` for some phaseII studies, I found that the centor of mass energy is not changable via a kwarg as luminosity. Therefore, I made this modification to enable the change of center of mass energy.

### DIFF
--- a/src/mplhep/label.py
+++ b/src/mplhep/label.py
@@ -234,6 +234,7 @@ def exp_label(
     label="",
     year=None,
     lumi=None,
+    com = None,
     llabel=None,
     rlabel=None,
     fontname=None,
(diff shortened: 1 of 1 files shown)


src/mplhep/label.py
```diff
@@ -265,6 +266,7 @@ def exp_label(
             Year when data was collected
         lumi : float, optional
             Aggregate luminosity shown. Should require ``"data"`` to be ``True``.
+        com: float, optional, default is 13, but can be changed to 7/8/13.6/14 to fit differnt requirements
```
andrzejnovak, line 269:
```suggestion
        com: float, optional, default is 13, but can be changed to 7/8/13.6/14 to fit different requirements
```

andrzejnovak:
Thanks @DickyChant, looks like a reasonable addition. Just a usual reminder, the label can be overwritten entirely by `rlabel='stuff'`

DickyChant:
> Thanks @DickyChant, looks like a reasonable addition. Just a usual reminder, the label can be overwritten entirely by `rlabel='stuff'`

Many thanks for correcting things. Also thanks for the reminder, in fact I was using this feature until I realized that maybe I can help :)

PR #372, opened 2022-03-28
fix: bump black to 22.3.0 due to click 8.1 release

henryiii:
See https://github.com/psf/black/issues/2964 for details.

Committed via https://github.com/asottile/all-repos


andrzejnovak:
Thanks @henryiii

PR #379, opened 2022-04-13
fix(histplot): density to match mpl behaviour #378

### LINKED ISSUE #378: `density=True` does not account for bin edges
Currently the density calculation ignores the bin edges (it just computes the reciprocal sum of the counts) 

https://github.com/scikit-hep/mplhep/blob/146ab7f01499ec3547d4a8c96e1e421cf65a7239/src/mplhep/utils.py#L263

This probably should be something like `1/np.sum(np.diff(self.edges) * self.values)`

andrzejnovak:
- fix #378 
- Modify tests to meaningfully show the difference between density/binwnorm

@henryiii had pre-commit complain about 
```
ERROR: Could not find a version that satisfies the requirement numpy~=1.22.0
```
for 
```
- repo: https://github.com/pre-commit/mirrors-mypy
  rev: v0.942
  hooks:
  - id: mypy
    files: src
    additional_dependencies: [uhi~=0.3.0, numpy~=1.22.0]
```

Any thoughts on why that is?


henryiii:
I really need to bump that to actually being 1.22 (starting in boost-histogram). Guessing that's only showing up locally? UHI sets >= 1.13 publicly (though it's pre-commit does lock to 1.22.*, that can't be seen from an install)

andrzejnovak:
Possibly. Sorry I posted the .yaml snippet after the change. Indeed I didn't check if it would also error in the CI, by forcing past pre-commit. Is there any reason to actually require a specific minor version for numpy?

andrzejnovak:
Ok, I am gonna merge this as it now and see what happens with pre-commit.

PR #381, opened 2022-05-12
Used in CMS SMP-20-003

lmoureaux:
We used this awesome package for most plots in the recent CMS paper SMP-20-003. Add it to the list of papers in the `README` as it's good to see continued usage of the package and the previous entry is two years old.


andrzejnovak:
Cheers, great to hear, thanks!

PR #386, opened 2022-05-24
feat: add lumi_format option to exp_label

swertz:
Currently, the integrated luminosity is passed as float and is simply converted to a string for the lumi label, potentially resulting in a lot of decimal places being shown. Usually, one has to reduce the fixed floating point precision to a more reasonable one. 

Instead of letting the user round down the number manually before passing it, this adds a `lumi_format` option to `exp_label` to change how the number is formatted. The current default behaviour is unchanged.

Edit: an alternative would be to simply let the user pass the lumi as an already-formatted string. WDYT?

### DIFF
--- a/src/mplhep/label.py
+++ b/src/mplhep/label.py
@@ -236,6 +236,7 @@ def exp_label(
     label="",
     year=None,
     lumi=None,
+    lumi_format="{0}",
     com=None,
     llabel=None,
     rlabel=None,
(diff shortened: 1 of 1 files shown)


swertz:
Do you mean passing e.g. `"{0} $\mathrm{{fb^{{-1}}}}$"` and doing `.format(lumi)` on that? I dont't find that very convenient because of the mix of format and latex braces. Also, there is already the possibility to fully specify the `rlabel`, in case the user wants to deviate from the default `fb^-1` ...?

PR #388, opened 2022-06-01
fix: remote `r_align` from `__all__`

agoose77:
This change was originally made in #185


andrzejnovak:
Thanks for the cleanup @agoose77!

Issue #405, opened 2022-11-29
ENH: Gallery


jonas-eschle:
I think it would be useful to have a gallery of different plot styles for comparison and for users to choose from.

Should we add a "Gallery" in the docs? And then in the doc making create the plots?

andrzejnovak:
Indeed, I think this would be helpful, some examples are in the linked pyhep/agc demos, but it's not easily visible. I didn't have time yet to update to the newer style docs that eg hist/awkward have, but that would probably  be a good time to do it.

jonas-eschle:
Ah, you mean to use the pydata_sphinx_theme? Yeah, that could be a good occasion, but that can be a smooth transition.
What do you think, adding another top level menu entry named "Gallery"? Or anything else?

I can get the pydata theme up meanwhile

PR #406, opened 2022-11-29
docs: switch docs to pydata theme, add initial gallery, add some type checks

jonas-eschle:
Change to pydata and add a gallery top level as discussed in #405


andrzejnovak:
Hmm, the Gallery doesn't seem to show up in the docs preview

jonas-eschle:
Yep, not yet, it's still WIP, I fixed it now and added a small example (but at this point, I am not great with Sphinx tbh to create something nice)

jonas-eschle:
Hi @andrzejnovak , I've finally added something more useful to the gallery, it shows up and has two examples. It's probably not the most elegant way, but alas, better than nothing and gives a good idea of what is available.

Also fixed some typos here and there, the docs compile now without errors/warnings

jonas-eschle:
Okay, that was tricky with the path. Now sphinx complains, but it works. Before, it didn't complain and the pathes didn't work.

Better complaining and working than the other way around. Probably needs some better structure. But as said, it's there and works for now, provides a good overview in a simple manner.

Issue #337, opened 2021-09-21
Add an option to sum over/underflow with the adjacent visible bins


pieterdavid:
I didn't see any mention of this in the documentation, although it is reasonably common, and I think the plotting layer is the right place to do this (it seems more like a workaround elsewhere). Would a simple flag to `histplot` cover all common cases? Any preferences on naming? `show_overflow` is short, but may be found confusing... if I find a bit of time I'll try to implement this.

andrzejnovak:
I've been wondering about what to do with underflow/overflow for a while, especially what would be the right way to visualize it. I think we don't have to deal with the trivial case `histplot(h.to_numpy(flow=True)...)` but it would be good to have for `hist` or `TH1s` going in. Some thoughts I have are these:

- by default, there should be some indication if there is flow, maybe an arrow?
- I prefer shorter kwargs, `flow` seems to be used already in some context
- Seems to me we have 4 cases
  - Nothing
  - Indicator
  - Sum
  - Display a new bin (maybe with a gap from the rest of the histogram?)
- arguments could be
  - None/False
  - "hint" (default?)
  - "sum"/"add"
  - "show"/True

PR #413, opened 2023-03-02
feat : implement under/overflow function 

Ming-Yan:
workable version for #337 
- implemented in `histplot`,`hist2dplot` function
- "hint" :  show text if under/overflow exist, also printout the value 
-  "show": additional bin with 5 times larger bin width
-  "sum": sum under/overflow bin to first/last bin


andrzejnovak:
Thanks a lot! I'll take a look tomorrow.

andrzejnovak:
Alright. Can you split this into two PRs, I think the `plot_ratio` plot is for a separate discussion. For to flow bin options, it would be nice if you could post what they look like and add them as tests.

src/mplhep/plot.py
```diff
@@ -130,6 +131,8 @@ def histplot(
             Attempts to draw x-axis ticks coinciding with bin boundaries if feasible.
         ax : matplotlib.axes.Axes, optional
             Axes object (if None, last one is fetched or one is created)
+        flow :  str, optional {None, "show", "sum"}
```
andrzejnovak, line 134: `hint` option missing?

src/mplhep/plot.py
```diff
@@ -392,17 +429,45 @@ def iterable_not_string(arg):
...
+        if binticks:
+            _slice = int(round(float(len(final_bins)) / len(ax.get_xticks()))) + 1
+            ax.set_xticks(final_bins[::_slice])
     else:
         ax.set_xticks(_bin_centers)
         ax.set_xticklabels(xtick_labels)
 
     if x_axes_label:
         ax.set_xlabel(x_axes_label)
+    if flow == "hint":
```
andrzejnovak, line 446: I was thinking more along the lines of an arrow maybe? I am also not sure this should be printing (for example if one is making a batch of plots, this might just spam the console)

src/mplhep/plot.py
```diff
@@ -392,17 +429,45 @@ def iterable_not_string(arg):
...
+        underflow, overflow = 0.0, 0.0
+        for h in hists:
+            underflow = underflow + h.view(flow=True)["value"][0]
```
andrzejnovak, line 449: Is the intended behaviour for a stack of histogram a single flow bin?

src/mplhep/plot.py
```diff
@@ -482,6 +550,32 @@ def hist2dplot(
     H = hist.values()
     xbins, xtick_labels = get_plottable_protocol_bins(hist.axes[0])
     ybins, ytick_labels = get_plottable_protocol_bins(hist.axes[1])
+
+    # Show under/overflow bins
+    # "show": Add additional bin with 5 times bin width
+    if flow == "show":
```
andrzejnovak, line 556: I am a bit worried this will end up with weird proportions where the informative part of the hist is just a small thing in the centre with huge borders

src/mplhep/plot.py
```diff
@@ -571,8 +683,320 @@ def hist2dplot(
...
 
 
+### copy functions coffea.hist.plotratio https://github.com/CoffeaTeam/coffea/blob/master/coffea/hist/plot.py to boost-hist
+################
+## ratio plot ##
+################
+_coverage1sd = scipy.stats.norm.cdf(1) - scipy.stats.norm.cdf(-1)
+
+
+def plotratio(
```
andrzejnovak, line 693: This should be a separate PR, for discussion.

src/mplhep/plot.py
```diff
@@ -571,8 +683,320 @@ def hist2dplot(
...
+    if clear:
+        ax.autoscale(axis="x", tight=True)
+        ax.set_ylim(0, None)
+
+    return ax
+
+
 #############################################
 # Utils
+def compatible(self, other):
```
andrzejnovak, line 890: This should go into https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/utils.py

src/mplhep/plot.py
```diff
@@ -571,8 +683,320 @@ def hist2dplot(
...
+    if len(self.axes) != len(other.axes):
+        return False
+    if set(self.axes.name) != set(other.axes.name):
+        return False
+    # if not all(d1 == d2 for d1, d2 in zip(self.dense_axes(), other.dense_axes())):
+    #     return False
+    return True
+
+
+def poisson_interval(sumw, sumw2, coverage=_coverage1sd):
```
andrzejnovak, line 901: This is already in https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/error_estimation.py

src/mplhep/plot.py
```diff
@@ -571,8 +683,320 @@ def hist2dplot(
...
+        scale[missing] = scale[argnearest]
+    counts = sumw / scale
+    lo = scale * scipy.stats.chi2.ppf((1 - coverage) / 2, 2 * counts) / 2.0
+    hi = scale * scipy.stats.chi2.ppf((1 + coverage) / 2, 2 * (counts + 1)) / 2.0
+    interval = np.array([lo, hi])
+    interval[interval == np.nan] = 0.0  # chi2.ppf produces nan for counts=0
+    return interval
+
+
+def normal_interval(pw, tw, pw2, tw2, coverage=_coverage1sd):
```
andrzejnovak, line 944: These two should also go in here https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/error_estimation.py

src/mplhep/plot.py
Ming-Yan: As we already have option to display, then I just remove printing

src/mplhep/plot.py
Ming-Yan: yes, it's intended to just show stacked number, showing sum of the stacked bin

src/mplhep/plot.py
Ming-Yan: or we can move to 2 times bin width?

src/mplhep/plot.py
Ming-Yan: move to another PR together with ratioplot

Ming-Yan:
Add 1d/2d test plot 
test_histplot_flow
test_hist2dplot_flow

alexander-held:
Hi, thanks for working on this! The "default" option above draws a lot of attention on the outmost bins, but does not explain what the red color corresponds to. In my opinion there is no special visualization needed if flow content is not included: this is the only case where the plot actually does what it says on the axis.

The "show" and "sum" cases on the other hand are not doing this, they suggest the bin contents have a value of the observable within the range indicated by the bin edges (which is not necessarily true). I think it would be helpful to denote somehow that those are special. One way in which this could be accomplished is by breaking the axis in the middle of the flow (or sum) bin. In principle this would also require the axis labels to be updated accordingly. This might not generalize too well, and I don't know how to translate that to 2d.

Ming-Yan:
Hi @alexander-held , thanks for the suggestion, 
> Hi, thanks for working on this! The "default" option above draws a lot of attention on the outmost bins, but does not explain what the red color corresponds to. In my opinion there is no special visualization needed if flow content is not included: this is the only case where the plot actually does what it says on the axis.

Indeed this is the "hint" option, where highlight the bins if under/overflow contents exist

> The "show" and "sum" cases on the other hand are not doing this, they suggest the bin contents have a value of the observable within the range indicated by the bin edges (which is not necessarily true). I think it would be helpful to denote somehow that those are special. One way in which this could be accomplished is by breaking the axis in the middle of the flow (or sum) bin. In principle this would also require the axis labels to be updated accordingly. This might not generalize too well, and I don't know how to translate that to 2d.

To have break axis, do you mean the style like this?https://matplotlib.org/3.1.0/gallery/subplots_axes_and_figures/broken_axis.html
I agree with the point that the flow bins value not next to the bin edges, but I think we also see some plots using the current implementation. For the sum option, as it merges to the last/first bin, I think split axis is not required

alexander-held:
> Indeed this is the "hint" option, where highlight the bins if under/overflow contents exist

Thanks for clarifying, I misunderstood this then. If the default behavior is that the plots do not change at all then that sounds reasonable to me.

For the broken axis: yes, I was thinking of something like that. To be clear here: I don't think this is crucially needed, it was just meant as an idea. I know that in practice there are many plots that use e.g. the "sum" approach. This is not specific to the changes here, it is a more general comment. Having the bins end at a clearly defined value on the axis that does not actually correspond to the cutoff for the bin content is what I find personally misleading. It is even more misleading in the examples above where the axis continues to the left and the right of the histogram, suggesting additional empty bins (so really suggesting that all events are in [5, 15]). I am not opposed to this implementation here though, users are still free to use what they want of course.

andrzejnovak:
Thanks a lot @Ming-Yan for adding the pics. I am +1 on the broken axis (particularly for the "hint" and "show" options). 

>  In my opinion there is no special visualization needed if flow content is not included: this is the only case where the plot actually does what it says on the axis.

I am not sure I agree that ignoring the flow bin does actually "do what it says" because for most plots the xlims don't match the extent of the histogram, but include some edge whitespace, which would suggest there's nothing there, rather than "no info here" which could be well shown with the broken axis.

alexander-held:
Agreed on the issue of xlim not matching the range of the histogram. I guess I'd always set the limits so that this cannot happen in practice.

Ming-Yan:
Thanks  for the discussion, I would try to implemented broken axis for "hint" and "show" option

NJManganelli:
+1 for Broken axis to show flow bins, this seems like a really nice way to communicate it visually

Ming-Yan:
Implemented broken axis signed(not using subplots)
Also edit the axis label for `show` option

image
image

andrzejnovak (commented):
Nice. I think this is definitely in the right direction. The added test baseline png is not updated and the existing test baseline pngs should not be changed in this PR (there's no visual difference, it's just that they're not bit by bit exact copies, so doing `git commit -a` is adding them on. 

For the viz in 1D I think we don't need the markers on the top axis and I think it would look better if the "broken axis" markers didn't go through the bin edges, but were contained "within" the bin. 

I think if the gap value in `"show"` is set to nan it should also help in the 2D plot, where right now it looks like a 0 value bin, as opposed to a "missing" one.

src/mplhep/plot.py
Ming-Yan: If it's the `show` option, it stacked like other bins.

src/mplhep/plot.py
```diff
@@ -168,9 +171,42 @@ def histplot(
...
+
     plottables = [
         Plottable(h.values(), edges=final_bins, variances=h.variances()) for h in hists
     ]
+    # Show under/overflow bins
+    # "show": Add additional bin with 5 times bin width
+    if flow == "show":
+        plottables = []
+        final_bins = np.array(
+            [
```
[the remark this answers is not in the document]
Ming-Yan: Try to implemented suggested case with 3%, 5% with unequal bin case, visually looks good to me
image
image

src/mplhep/plot.py
Ming-Yan: add new test plots 
image

Ming-Yan:
Thanks @andrzejnovak  for reviewing the updates. The suggests implemented in the new commit with adding a few more tests

src/mplhep/plot.py
```diff
@@ -168,9 +171,42 @@ def histplot(
     )
 
     # Cast to plottables
+
```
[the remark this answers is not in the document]
andrzejnovak: Still present?

src/mplhep/plot.py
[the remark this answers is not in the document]
andrzejnovak: Thanks, this looks good to me.

src/mplhep/plot.py
```diff
@@ -168,9 +171,42 @@ def histplot(
...
+            [
+                final_bins[0] - (final_bins[1] - final_bins[0]) * 3,
+                final_bins[0] - (final_bins[1] - final_bins[0]),
+                *final_bins,
+                final_bins[-1] + (final_bins[1] - final_bins[0]),
+                final_bins[-1] + (final_bins[1] - final_bins[0]) * 3,
+            ]
+        )
+        for h in hists:
+            value, variance = h.view(flow=True)["value"], h.view(flow=True)["variance"]
```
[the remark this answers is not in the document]
andrzejnovak: Is this implemented? The code still seems to insert 0s' as opposed to Nans, though the gap bins no longer show error bar on the updated viz in another comment. What happens if you `set_ylim(-1, None)` there should ideally be nothing there in the gap

src/mplhep/plot.py
```diff
@@ -536,6 +694,76 @@ def hist2dplot(
         cb_obj = None
 
     plt.sca(ax)
+    if flow == "hint" or flow == "show":
+        d = 0.9  # proportion of vertical to horizontal extent of the slanted line
```
andrzejnovak, line 698: Could the size of the broken axis marker scale of the figsize?

andrzejnovak:
@Ming-Yan pre-commit should now work correctly

Ming-Yan:
Adapt two-side, one-side no overflow bins plot
image

tests/test_basic.py
```diff
@@ -110,6 +111,112 @@ def test_histplot_density():
...
+
+    f = uproot.recreate("flow_th1.root")
+    f["h"] = h
+    f["h2"] = h2
+    f["h3"] = h3
+    f["h4"] = h4
+    fig, axs = plt.subplots(2, 2, sharey=True, figsize=(10, 10))
+    axs = axs.flatten()
+
+    hep.histplot(h, ax=axs[0], flow="show")
```
andrzejnovak, line 155: This doesn't need to get created when testing `hist`. Also instead of the bins existing and you not filling them, can you replace it with 
```
h = hist.Hist(hist.axis.Regular(20, 5, 15, name="x", flow=True), hist.storage.Weight())
h2 = hist.Hist(hist.axis.Regular(20, 5, 15, name="x", underflow=True, overflow=False), hist.storage.Weight())
h3 = hist.Hist(hist.axis.Regular(20, 5, 15, name="x", underflow=False, overflow=True), hist.storage.Weight())
h4 = hist.Hist(hist.axis.Regular(20, 5, 15, name="x", flow=False), hist.storage.Weight())
```

These can also iirc be shortened with `hist.new.Reg(20, 5, 15, name="x", underflow=True, overflow=False).Weight()`

tests/test_basic.py
```diff
@@ -110,6 +111,112 @@ def test_histplot_density():
...
+    f = uproot.open("flow_th1.root")
+    h = f["h"]
+    h2 = f["h2"]
+    h3 = f["h3"]
+    h4 = f["h4"]
+
+    fig, axs = plt.subplots(2, 2, sharey=True, figsize=(10, 10))
+    axs = axs.flatten()
+
+    hep.histplot(h, ax=axs[0], flow="show")
```
andrzejnovak, line 184: This still uses the hist object no? I am not sure if just passing f["h"] will pass the TH1 or if you need to close and load the file, can you check?

tests/test_basic.py
```diff
@@ -110,6 +111,112 @@ def test_histplot_density():
...
+    histh = hist.Hist(
+        hist.axis.Regular(20, 5, 15, name="x", flow=False), hist.storage.Weight()
+    )
+    nph, bins = np.histogram(entries, bins=20, range=(5, 15))
+    histh.fill(entries)
+
+    fig, axs = plt.subplots(1, 2, sharex=True, sharey=True, figsize=(10, 5))
+    axs = axs.flatten()
+
+    hep.histplot(histh, ax=axs[0], flow="hint")
```
andrzejnovak, line 212: Could set `yerr=False` such that both look the same.

tests/test_basic.py
```diff
@@ -110,6 +111,112 @@ def test_histplot_density():
...
+def test_histplot_flow():
+    np.random.seed(0)
+    h = hist.Hist(hist.axis.Regular(20, 5, 15, name="x"), hist.storage.Weight())
+    h.fill(np.random.normal(10, 3, 400))
+    fig, axs = plt.subplots(1, 3, sharey=True, figsize=(15, 5))
+    axs = axs.flatten()
+
+    hep.histplot(h, ax=axs[0], flow="hint")
+    hep.histplot(h, ax=axs[1], flow="show")
+    hep.histplot(h, ax=axs[2], flow="sum")
```
andrzejnovak, line 124: Add the `flow=None` test back in?

src/mplhep/plot.py
```diff
@@ -4,6 +4,7 @@
 from collections import OrderedDict, namedtuple
 from typing import TYPE_CHECKING, Any, Union
 
+import hist as Hist
```
andrzejnovak, line 7: is there a reason this needs to be renamed?

src/mplhep/plot.py
```diff
@@ -167,11 +172,85 @@ def histplot(
...
+            underflow = underflow + h.values(flow=True)[0]
+            overflow = overflow + h.values(flow=True)[-1]
+
+    # "show": Add additional bin with 5 times bin width
+    plottables = []
+    flow_bins = final_bins
+    for i, h in enumerate(hists):
+        value, variance = h.values(), h.variances()
+        if not isinstance(h, Hist.hist.Hist) and not "uproot.models.TH" in str(type(h)):
+            print(f"Warning: {type(h)} is not allowed to get flow bins")
```
andrzejnovak, line 201: Hmm, is this triggered every time you pass just an array as a hist?

src/mplhep/plot.py
```diff
@@ -167,11 +172,85 @@ def histplot(
...
+            flow = None
+            plottables.append(Plottable(value, edges=final_bins, variances=variance))
+        # check the original hist as flow bins
+        elif isinstance(h, Hist.hist.Hist) and len(h.values(flow=True)) < len(
+            final_bins
+        ):
+            print(f"Warning:  you don't have flow bins stored in {h}")
```
andrzejnovak, line 208: Similar as above

src/mplhep/plot.py
```diff
@@ -397,12 +476,69 @@ def iterable_not_string(arg):
...
         ax.set_xticklabels(xtick_labels)
 
     if x_axes_label:
         ax.set_xlabel(x_axes_label)
+    if flow == "hint" or flow == "show":
+        d = 0.9  # proportion of vertical to horizontal extent of the slanted line
+        trans = mpl.transforms.blended_transform_factory(ax.transData, ax.transAxes)
+        kwargs = dict(
+            marker=[(-0.5, -d), (0.5, d)],
+            markersize=15,
```
andrzejnovak, line 494: could the marker size be somehow derived in relation to figure size?

src/mplhep/plot.py
andrzejnovak: Seems you have this solved for 2D with `and flow is not None`

andrzejnovak (changes requested):

src/mplhep/plot.py
Ming-Yan: `hist` is used [here](https://github.com/scikit-hep/mplhep/blob/199430685ff73788a942ab9862fd9590a258f297/src/mplhep/plot.py#L618)

andrzejnovak: It would be better to rename that instead

tests/test_basic.py
Ming-Yan: I guess we don't need `None` anyway?

tests/test_basic.py
Ming-Yan: Hmm... the suggestion works. but if we convert the hist functions to ROOT TH1 the functions not act correctly...
The bins and value would mismatch for one-sided case

tests/test_basic.py
Ming-Yan: No, this use numpy hist

Ming-Yan: passing f["h"] will pass the TH1

src/mplhep/plot.py
Ming-Yan: This should be triggered every time, otherwise the hist would not filled as `flow` set to `None` after 1 iteration

src/mplhep/plot.py
Ming-Yan: similar as above

src/mplhep/plot.py
Ming-Yan: yes, it possible, I take height to change the size

```
ax_h = ax.bbox.height
        kwargs = dict(
            marker=[(-0.5, -d), (0.5, d)],
            markersize=ax_h*0.05,
            linestyle="none",
            color="k",
            mec="k",
            mew=1,
            clip_on=False,
            transform=trans,
        )
```

src/mplhep/plot.py
andrzejnovak: Is this really a good practice tho? I don't think I should see this warning if I all I do is call `histplot([1,2,3,2,1])` which I think will happen here right?

tests/test_basic.py
andrzejnovak: Yeah, this won't work for TH1 since ROOT cannot properly store hists without flow bins

Ming-Yan: so should I keep the previous structure or moving to the `hist` suggestion here?

src/mplhep/plot.py
Ming-Yan: Right, add `and flow is not None` and indeed if we have `None` option, we would keep the case `plottables.append(Plottable(value, edges=final_bins, variances=variance))`

tests/baseline/test_histplot_flow.png
andrzejnovak, line 1: This doesn't look right for `None`

tests/baseline/test_histplot_hist_flow.png
andrzejnovak, line 1: The one-sided flow tests seem to misbehave as well

andrzejnovak (changes requested):

src/mplhep/plot.py
```diff
@@ -167,11 +172,91 @@ def histplot(
...
-        Plottable(h.values(), edges=final_bins, variances=h.variances()) for h in hists
-    ]
+    # Show under/overflow bins
+    ## check underflow/overflow bin exist
+
+    underflow, overflow = 0.0, 0.0
 
+    for h in hists:
+        if (
+            not isinstance(h, hist.hist.Hist)
```
andrzejnovak, line 182: Also, can you check for `hist` type in the same way as you do for TH1? We don't need to make `hist` a dependency.

andrzejnovak: Actually, perhaps it would be best to just check it functionally aka 
```
hasattr(h, 'values') and 'flow' in inspect.getfullargspec(h.values).args
```
which should work for both `hist` and `TH1` or anything random that will conform the `uhi` spec (https://uhi.readthedocs.io/en/latest/)

tests/test_basic.py
```diff
@@ -2,6 +2,7 @@
 
 import os
 
+import hist
```
andrzejnovak, line 5: Gotta add it to test reqs as well https://github.com/scikit-hep/mplhep/blob/master/setup.py#L16

tests/baseline/test_histplot_flow.png
Ming-Yan: image
find a weird bug, not sure where it from, if I swap the order of "sum" and None then I get the correct version, but we do this independently...

tests/baseline/test_histplot_hist_flow.png
Ming-Yan: Hmmm.. I think the problem is you cannot check the flow bin is underflow or overflow from the stored `hist`... You would see there's a flow bins(no `h.values(underflow=True)` to check)

`Hist(Regular(20, 5, 15, overflow=False, name='x'), storage=Weight()) # Sum: WeightedSum(value=359, variance=359) (WeightedSum(value=377, variance=377) with flow)`

Ming-Yan: ah ok, then i need to add additional check for under/overflow bins for `hist` case, I just simply check

```
underflow = underflow + h.values(flow=True)[0]
overflow = overflow + h.values(flow=True)[-1]
```

tests/baseline/test_histplot_flow.png
andrzejnovak: So I checked out the PR, it seems the code is modifying the input object at some point, that's why it makes a mess
image

Ming-Yan: but we don't have any replacement either on `H` the input object and the plottable `h`, we only edit the value and variances. I don't get the point why the hist object would be modified..

Ming-Yan: Ah.. sorry I think it's the confusion of check [under/overflow bin](https://github.com/Ming-Yan/mplhep/blob/master/src/mplhep/plot.py#L191-L193 ) where the `elif flow is None : continue` should be added. And I don't think it changes the hist object(we didn't modify it anywhere in the code)

src/mplhep/plot.py
```diff
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import collections.abc
+import inspect
 from collections import OrderedDict, namedtuple
 from typing import TYPE_CHECKING, Any, Union
 
+import hist
 import matplotlib as mpl
```
andrzejnovak, line 9: This is now extra

src/mplhep/plot.py
```diff
@@ -167,11 +173,98 @@ def histplot(
...
+                overflow = overflow + h.values(flow=True)[0]
+        else:
+            underflow = underflow + h.values(flow=True)[0]
+            overflow = overflow + h.values(flow=True)[-1]
+
+    # "show": Add additional bin with 5 times bin width
+    plottables = []
+    flow_bins = final_bins
+    for i, h in enumerate(hists):
+        if isinstance(h, hist.hist.Hist):
```
andrzejnovak, line 203: check this via `hasattr`s as well

src/mplhep/plot.py
```diff
@@ -167,11 +173,98 @@ def histplot(
...
+            hasattr(h, "values")
+            and not "flow" in inspect.getfullargspec(h.values).args
+            and flow is not None
+        ):
+            if flow == "sum" or flow == "show":
+                print(f"Warning: {type(h)} is not allowed to get flow bins")
+            flow = None
+            plottables.append(Plottable(value, edges=final_bins, variances=variance))
+        # check the original hist as flow bins
+        elif isinstance(h, hist.hist.Hist) and len(h.values(flow=True)) < len(
```
andrzejnovak, line 217: check this via hasattrs as well

src/mplhep/plot.py
```diff
@@ -476,22 +630,88 @@ def hist2dplot(
...
+        hasattr(h, "values")
+        and not "flow" in inspect.getfullargspec(h.values).args
+        and flow is not None
+    ):
+        print(
+            f"Warning: {type(h)} is not allowed to get flow bins, flow bin option set to None"
+        )
+        flow = None
+    elif (
+        isinstance(h, hist.hist.Hist)
```
andrzejnovak, line 652: check this via hasattrs as well

andrzejnovak (changes requested):
There are still couple of explicit `hist` checks to be rid of in the code.  Plus some flake8 errors https://results.pre-commit.ci/run/github/184555939/1681661620.370AvrXARZSbUqgBkBfvJQ

andrzejnovak:
Thanks a lot @Ming-Yan!

src/mplhep/plot.py
```diff
@@ -476,22 +635,104 @@ def hist2dplot(
...
+        flow = None
+    elif (
+        hasattr(h, "axes")
+        and hasattr(h.axes[0], "traits")
+        and hasattr(h.axes[0].traits, "underflow")
+        and not h.axes[0].traits.underflow
+        and not h.axes[0].traits.overflow
+    ):
+        flow = None
+        print(f"Warning:  you don't have flow bins stored in {h}")
```
henryiii, line 664: This should have been:

```python
warnings.warn(f"You don't have flow bins stored in {h!r}")
```

henryiii:
Also, I don't understand the default. Calling plot on a flowless histogram should not produce a warning about missing flow bins; you should not have to pass `flow=None` to get it to shut up. Also, if it takes a text string, it should be possible to specify everything via a text string, and None should be the same as not specifying anything. That's the standard meaning of None, and the reason it's typed as "Optional".

Ming-Yan:
Apologize for noticing the suggestions late and thanks for the fixes

Issue #26, opened 2019-12-10
Enable error bands for histograms


andrzejnovak:

andrzejnovak:
Hi @hamzahanif2210, this is not currently implemented. If you'd like to work on the feature, I'd ask you to give a quick sketch here on what API and look you'd imagine for it.

danielhundhausen:
Hi @andrzejnovak , I'd be happy to work on this. As a standalone function something like this would work:

```
def errorband(
    H_band,  # value array if error band is symmetric / tuple of upper and lower value arrays
    bins,  # corresponding to bin edges of the hist values supplied in H_band
    label=None,
    edges=True,
    ax=None,
    **kwargs,
):
    # ax check
    if ax is None:
        ax = plt.gca()
    else:
        if not isinstance(ax, plt.Axes):
            raise ValueError("ax must be a matplotlib Axes object")

    # Construction of upper and lower boundries
    if isinstance(H_band, tuple):
        y_upper = H_band[0]
        y_lower = H_band[1]
    if isinstance(H_band, np.ndarray):
        y_upper = H_band
        y_lower = -y_upper

    error_band_args = { 
        "edges": bins, "facecolor": "none", "linewidth": 0,
        "alpha": .9, "color": "black", "hatch": "///"
    }   
    ax.stairs(y_upper, baseline=y_lower, label=label, **{**error_band_args, **kwargs})
```

Please let me know your thoughts.

andrzejnovak:
Welcome, @danielhundhausen! I think I was originally thinking that this could be an option in `histplot` along the line of `histplot(histtype='band')`, where the band could be determined either by setting `yerr` or automatically from the existing weights methods. Is there a reason why you'd say it would be better as a separate function?

danielhundhausen:
Thank you for your comments @andrzejnovak ! I made a PR #414 that includes the integration of errorbands in `histplot` when plotting with `histtype="fill"`, which is the standard use case, I think.
Regarding your question about making a separate function for the plotting - I think it reduces complexity of the code to split things into separate functions whenever possible. For the integration in `histplot` I have tried to follow the pattern of `errorbar`.

PR #414, opened 2023-03-21
feat: Add errorband plotting to histplot

danielhundhausen:
Adds a function to plot errorbands as well as integration of errorbands into `histplot` when plotting with `histstyle="fill"` and specifying `yerr`, addressing #26 .


andrzejnovak:
Sorry @danielhundhausen that we never converged on this. Please check the new release with #472 to see if that works for your usecase as well.

PR #416, opened 2023-04-03
feat : add ratioplot 

Ming-Yan:
- add ratioplot  function originate from [coffea.hist.plotratio](https://github.com/CoffeaTeam/coffea/blob/master/coffea/hist/plot.py)
- add  test function

### DIFF
--- a/src/mplhep/__init__.py
+++ b/src/mplhep/__init__.py
@@ -19,6 +19,7 @@ from .plot import (
     histplot,
     make_square_add_cbar,
     mpl_magic,
+    ratioplot,
     rescale_to_axessize,
     sort_legend,
     ylow,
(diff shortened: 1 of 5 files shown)


jonas-eschle:
hi @Ming-Yan what's the status of this? Can you maybe rebase the PR?

Ming-Yan:
Hi @jonas-eschle sorry for the long delayed, indeed I only made the ratio plot available for the boost hist object only but did not have time to investigate further. I can do rebase for the changes.

jonas-eschle:
The ci currently fails, there seems to be still issues, can you have a look at it? Also, is there a discussion/issue somewhere where the rationals of the functions are discussed? I.e. interface, defaults?

PR #426, opened 2023-05-15
fix: flow warning fix

henryiii:
This is the full suggested fix to the issues introduced in #413:

* Made the output a warning
* Used the repr of the histogram instead of the print (which is a textual plot)
* Avoided printing the warning if "none"/None is set for `flow=`.

Personally, I'd recommend not supporting `None`, and only supporting `"none"`, since it's just a mode like the others and it's not the default if no argument is passed.

I think "hint" should also not print a warning.

### DIFF
--- a/src/mplhep/plot.py
+++ b/src/mplhep/plot.py
@@ -509,7 +509,8 @@ def histplot(
 
     if x_axes_label:
         ax.set_xlabel(x_axes_label)
-    if flow == "hint" or flow == "show":
+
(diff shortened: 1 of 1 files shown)


henryiii:
What does hint look like? - I couldn't tell from the PR as it kept changing.

andrzejnovak:
Thanks for the fix! Actually thought we settled on no warn for hint if flow bins don't exist, but seems sth went wrong there.

Issue #431, opened 2023-06-12
[Bug] histplot raises IndexError when plotting an histogram with one bin


tcuisset:
Hi,
`histplot` crashes when plotting histograms with a single bin, when trying to draw overflow bins.

## Reproducer :
```python
import hist
import mplhep as hep

h = hist.Hist(hist.axis.Regular(1, 0, 1))
h.fill([-1, 0.5])
hep.histplot(h)
```
## Output
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[17], line 1
----> 1 hep.histplot(h)

File [***/lib/python3.10/site-packages/mplhep/plot.py:532], in histplot(H, bins, yerr, w2, w2method, stack, density, binwnorm, histtype, xerr, label, sort, edges, binticks, ax, flow, **kwargs)
    528 if underflow > 0.0:
    529     if flow == "hint":
    530         ax.plot(
    531             [
--> 532                 final_bins[0] - (final_bins[-3] - final_bins[2]) * 0.03,
    533                 final_bins[0],
    534             ],
    535             [0, 0],
    536             **kwargs,
    537         )
    538     if flow == "show":
    539         ax.plot(
    540             [flow_bins[1], flow_bins[2]],
    541             [0, 0],
    542             **kwargs,
    543         )

IndexError: index -3 is out of bounds for axis 0 with size 2
```

andrzejnovak:
Thanks for the report @tcuisset, @Ming-Yan could you take a look?

PR #432, opened 2023-06-13
fix : flow plot bug on axis range for histplot and hist2dplot

Ming-Yan:
change the axis range to fix the bug #431


andrzejnovak:
Thanks for the super quick fix. Should this affect the existing baseline images? Also could you add a quick test, based on the example in the issue? Perhaps using sth like `@check_figures_equal` instead of generating a new baseline image test like here: https://github.com/scikit-hep/mplhep/blob/bd907053e3a56473e9e44fd0751c8a95350b9108/tests/test_inputs.py#L56

Ming-Yan:
Indeed the split of the broken axis ticks slightly change in this PR. Regarding to the new test based on this issue, I am a bit confused what to compared in this test

Ming-Yan:
Indeed I find out the plot could be very different due to the numpy hist not able to have flow bins. create new test with this issue
image

PR #433, opened 2023-07-12
Update label.py

ruelasvh:
Remove hard-coded lumi for loc==4


andrzejnovak:
Thanks @ruelasvh. Can you also mod the failing test such that it shows the behaviour when lumi number is supplied?

ruelasvh:
> Thanks @ruelasvh. Can you also mod the failing test such that it shows the behaviour when lumi number is supplied?

Hi @andrzejnovak I've updated the MR. The needed to fix for the default 13 TeV com. Now the tests should be passing.

andrzejnovak:
Hi @ruelasvh, sorry this took a while to get in, seems we just needed to update the modified test. I'll mint a release for you.

Issue #436, opened 2023-08-07
TypeError when plotting 2d histograms with flow="show"


alexander-held:
Plotting a 2d histogram with `flow="show"` results in errors like this:

> TypeError: Dimensions of C (12, 12) should be one smaller than X(11) and Y(11) while using shading='flat' see help(pcolormesh)

I am using `mplhep==0.3.28`.

Reproducer:
```python
import hist
import numpy as np

h = hist.Hist.new.Regular(10, 0, 1, name="x").Regular(10, 0, 1, name="y").Double()
h.fill(x=np.random.random(100), y=np.random.random(100))
h.plot(flow="show")
```

Full trace:
```pytb
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[25], line 6
      4 h = hist.Hist.new.Regular(10, 0, 1, name="x").Regular(10, 0, 1, name="y").Double()
      5 h.fill(x=np.random.random(100), y=np.random.random(100))
----> 6 h.plot(flow="show")

File [...]/lib/python3.9/site-packages/hist/basehist.py:400, in BaseHist.plot(self, overlay, *args, **kwargs)
    397     return self.plot1d(*args, overlay=overlay, **kwargs)
    399 if self.ndim == 2:
--> 400     return self.plot2d(*args, **kwargs)
    402 raise NotImplementedError("Please project to 1D or 2D before calling plot")

File [...]/lib/python3.9/site-packages/hist/basehist.py:452, in BaseHist.plot2d(self, ax, **kwargs)
    446 """
    447 Plot2d method for BaseHist object.
    448 """
    450 from hist import plot
--> 452 return plot.hist2dplot(self, ax=ax, **_proc_kw_for_lw(kwargs))

File [...]/lib/python3.9/site-packages/mplhep/plot.py:753, in hist2dplot(H, xbins, ybins, labels, cbar, cbarsize, cbarpad, cbarpos, cbarextend, cmin, cmax, ax, flow, **kwargs)
    750 X, Y = np.meshgrid(xbins, ybins)
    752 kwargs.setdefault("shading", "flat")
--> 753 pc = ax.pcolormesh(X, Y, H, vmin=cmin, vmax=cmax, **kwargs)
    755 if x_axes_label:
    756     ax.set_xlabel(x_axes_label)

[...]

PR #437, opened 2023-08-08
fix: overflow bug 

Ming-Yan:
Fix the bug mentioned in #436 

Under/overflow bins define as non-zero value


andrzejnovak:
Hi @Ming-Yan, thanks for the quick fix and sorry about the delay in taking a look. Could you double-check this and make sure we test across the matrix of options of the 4 flow parameters and different combinations of possible hist objects?

I am currently getting a shape error with the following MRE
```
h = hist.new.Reg(10, 0, 1, name='x').Var([0,1,2,3], name='y').Weight()
h.fill(np.random.normal(0.5, 0.5, 1000), np.random.choice([0,1,2,3], 1000))
h.plot(flow='show')
```

Ming-Yan:
Hi @andrzejnovak , thanks for the feedback, indeed the padding `nan` axis and bins are not correctly implemented in the previous iteration, now fix in this update. 
I also check the failure on the CI pipeline(it's good with my local test with the command(also CI in python 3.8) but still some error for 3.7...
`python -m pytest -r sa --mpl --mpl-results-path=pytest_results`

andrzejnovak:
Did the existing baseline images change? Perhaps the updated version is missing from the PR?

Ming-Yan:
Hmmm... no, the local test only has problem with lhcb font style...

andrzejnovak:
I am a bit confused, the diff you linked seems to be for the flow test right? I'd expect the changes not affect the LHCb style here so if it's failing locally it shouldn't be an issue (you're probably missing msft fonts - LHCb uses Times New Roman).

If it'd help, we're gonna drop python 3.7 anyway in #428 so I can merge that first if you'd want to rebase on that.

Ming-Yan:
> I am a bit confused, the diff you linked seems to be for the flow test right? I'd expect the changes not affect the LHCb style here so if it's failing locally it shouldn't be an issue (you're probably missing msft fonts - LHCb uses Times New Roman).

Indeed the first plot(white background) is the local test, where no issue with flow plot but missing font in the machine. The second test is from the CI pipeline where the failure of `test_hist2dplot_flow` appears.

> If it'd help, we're gonna drop python 3.7 anyway in #428 so I can merge that first if you'd want to rebase on that.

Yeah, I think this would be helpful🙂

andrzejnovak:
> Yeah, I think this would be helpful🙂

alright, try now

andrzejnovak:
Superseded in #487

PR #447, opened 2023-10-03
feat: added colors parameter to allow user to select the colour of ea…

jjhw3:
Added a 'colors' parameter to histplot to allow users to set histogram colours in the same way as the labels. Before this change, one could not pass one color per histogram without histplot crashing, as all the colors were being passed into ax.stairs as a _kwargs entry.


andrzejnovak:
Hey @jjhw3 welcome. Could you describe the issue a bit more? I seem to easily be able to write 
```
hs = [hist.new.Reg(10,0,10).Weight().fill(np.random.normal(3+i, 1, 1000)) for i in range(3)]
hep.histplot(hs, histtype='fill', stack=True, color=['red', 'blue', 'firebrick']);
```
to get the following 
image

PR #448, opened 2023-10-03
fix: _chunked_kwargs not expanding kwargs if they're tuples even if t…

jjhw3:
fix: _chunked_kwargs not expanding kwargs if they're tuples even if the tuple isn't a valid mpl color.

This issue confused me for a long time and is likely to come up if one uses the 'zip' builtin to zip up colours and histograms. Although this could be fixed by simply converting the tuple of colors returned by zip into a list, this is a better solution as it does not require users to read the source code to figure this out.

### DIFF
--- a/src/mplhep/plot.py
+++ b/src/mplhep/plot.py
@@ -306,7 +306,9 @@ def histplot(
         # Check if iterable
         if iterable_not_string(kwargs[kwarg]):
             # Check if tuple of floats or ints (can be used for colors)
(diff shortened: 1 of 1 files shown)


jjhw3:
Before this fix, the following would not work (note the tuple of colors), not it does:

hs = [hist.new.Reg(10,0,10).Weight().fill(np.random.normal(3+i, 1, 1000)) for i in range(3)]
hep.histplot(hs, histtype='fill', stack=True, color=('red', 'blue', 'firebrick'));

andrzejnovak:
Ah, looks good. Thanks for the fix!

PR #449, opened 2023-10-03
fix: make passing colors as tuple safe

jjhw3:
Hi, sorry to raise this again, but it occurred to me that the previous solution would have nasty effects if you passed a generator in as an argument as it would loop through all the items in the generator. This is a safer implementation thanks to short-circuit evaluation.


andrzejnovak:
Hah, somehow I don't think that happens often, but thanks for catching it.

PR #453, opened 2023-10-04
feat(histplot): Add "mixed" histtype for stacked histograms

rpsimeon34:
feat(histplot): Add "mixed" option for histtype to allow some filled and one unfilled stacked histogram in one plot


andrzejnovak:
Welcome @rpsimeon34. There's for the PR, but could you walk me through a bit why this should be an included functionality that cannot be obtained otherwise? At first glance this seems very use case specific and not playing super nicely with things like passing the order kwarg.

rpsimeon34:
Hi @andrzejnovak, I actually wasn't aware that you could pass 'none' as an element in `facecolor`. The strongest argument I can think of now for including this as a feature is that this allows you to leverage matplotlib's automatic color assignment when plotting a variable-length list of histograms (as opposed to manually naming the colors to include). But, if you think this PR causes more difficulties than it solves, I'm happy to use the `facecolor` solution instead.

andrzejnovak:
Sorry for the long delay in replying and thanks for the discussion. I am going to close this for now. Please feel free to open it as an issue later on if the need for this functionality persists and can be argued as an at least somewhat common use case.

PR #455, opened 2023-10-11
fix: support older matplotlib

agoose77:
#443 broke support for older matplotlib (within our version bounds). This PR adds a version check.


andrzejnovak:
Yikes, thanks a lot for the fix!

PR #466, opened 2024-02-08
feat: Adding publication/supplemental info to label

yimuchen:
- First running version of adding publication information to the plot labels. (Works decently well for `loc=2`)
- Additional adjustments will need to done for all possible `loc`/`data` combinations

test_loc2_dataFalse
test_loc2_dataTrue


yimuchen:
Right, still a bunch of work needs to be done for many configurations... Will need quite a bit of tweaking.

test_pub_loc

yimuchen:
Updated for decent 1,2,3 settings, Not entirely sure what to do for 0 and 4 for the very extreme cases.
test_pub_loc

andrzejnovak:
Thanks! I think for loc=4 the SuppText should go on a new line like in loc=2. For loc=0 I am not sure. Perhaps in-frame or the right edge where the z-axis/cbar label would be.

yimuchen:
Updated for loc=0 and 4. Made sure that changes to 4 did not change the result without the arXiv ID.

test_label_loc
test_pub_loc

yimuchen:
I just checked with a friend in ALTAS, I don't think they have a prefered location for arXiv labels (and if they do, I guess they can contribute to change this). So let me know if I should do any additional adjustments to these current defaults.


test_pub_loc
test_label_loc

andrzejnovak (commented):
Sorry for the delay. Implementation-wise looks good to me. For `loc=0` I was rather thinking this could live along the y-axis, outside of the figure like this:
image

I'd also say the spacing in loc 1,2,3 is a bit tight (in 2,3 this can be seen in the test with some symbols overlaping).

For loc my preference would probably be for the SuppText to be below the xsec information.

What do you think?

yimuchen:
@andrzejnovak 
What do you think about this new tweaked result?
test_pub_loc

andrzejnovak:
Thanks a lot! I have a last comment, for loc=0 the orientation of the supptext should be flipped (such that if a person turns their head left, they can read both the supptext and y-axis label. Otherwise LGTM

yimuchen:
Done!
test_pub_loc

andrzejnovak:
@yimuchen it looks like the reference image is missing for the test

yimuchen:
Should I generate it from my side? Or should we have some CI process to generate these images to ensure consistency?

andrzejnovak:
The file should be generated locally and included in the tests/baseline. The CI will then check against it 

https://github.com/scikit-hep/mplhep/blob/master/CONTRIBUTING.md#generating-reference-visuals

andrzejnovak:
Thanks a lot @yimuchen. Let me mint a release.

PR #488, opened 2024-04-01
fix: array copy and logic for obtaining variances and values

jonas-eschle:
I've noticed the new `.copy()` to avoid that the underlying array changes (such as here https://github.com/scikit-hep/mplhep/blob/master/src/mplhep/plot.py#L191).

I would suggest to follow closer the [array protocol](https://numpy.org/doc/stable/user/basics.copies.html#copy), which uses the `np.copy` -> other, array-like types (like JAX, TensorFlow, PyTorch etc) may don't have the `copy` attribute but support the `np.copy` (as we've noticed in the [CI](https://github.com/zfit/zfit/actions/runs/8510783395/job/23310700551?pr=541#step:9:122)).

Also, the logic was possibly flawed, as a histogram also has `values` and therefore, it would take the second if, wouldn't it? Removed also the re-assignement of 0 to the flows, they were already initialized with 0.

Does the new flow makes sense?


jonas-eschle:
Looks fine from my side, mypy fails because of the over- and underflow being arrays, not floats. Feel free to adjust, I think it was like this before as well, or let me know what's the best way to go.

Otherwise can be merged from my side

andrzejnovak:
Huh, so the return type for `np.copy(float)` is apparently not `float` which is what was tripping up mypy. Calling copy on the array instead is probably a touch less efficient but works and we're not too concerned about performance for plotting. 

Can you check if this still fine in zfit? I'll merge it then.

jonas-eschle:
> Huh, so the return type for `np.copy(float)` is apparently not `float` which is what was tripping up mypy. Calling copy on the array instead is probably a touch less efficient but works and we're not too concerned about performance for plotting.

the type returned was actually a numpy.float64, but now it's an array with shape (), afaiu. The performance should be about the same, the expensive part is anyways the copy of large array (if at all), and that happens in both cases.

> Can you check if this still fine in zfit? I'll merge it then.

yes, works, can be merged

> We could also do `copy.copy` for these

that should also work I think, alternatively, we could also convert to floats... all works, as long as we don't use the `copy` attribute

andrzejnovak:
Alright, let's run with this for now, see if anything comes up laters. Thanks for the PR!

PR #494, opened 2024-04-24
ci: make python versions consistent, add other systems

jonas-eschle:


jonas-eschle:
@andrzejnovak ready from my side, just a bit of cleanup. The pre-commit cannot install mypy, it's probably a temporary failure.

PR #496, opened 2024-04-28
fix: legend linestyle for step histogram

8me:
The legend entries for `mplhep.histplot` for `histtype='step'` shows always continuous entires no matter what `linestyle` arg is defined. In order to fix that the linestyle is forwarded to `_e_leg` and `xerr` is set to `None` as it is not used in the step mode.


andrzejnovak:
Hi @8me thanks for the PR! I think I understand the change, but could you add a picture here just for posterity?

andrzejnovak:
Perfect, thanks for the PR!

Issue #501, opened 2024-06-02
Incorrect flow bin labeling


atownse2:
Here is a reproducer, using the main branch:
```python
import mplhep as hep
import hist

h = hist.Hist(
    hist.axis.Regular(4, 0.1, 0.9, underflow=True, overflow=True)
    ).fill(1*np.random.rand(1000))
hep.histplot(h, flow='show', label='h')

plt.show()
```
Which produces:
![4d271f4f-c26c-4af4-90a4-08d2db534d70](https://github.com/scikit-hep/mplhep/assets/47793576/3fd846a8-14e7-439d-a398-0fe1ba0d493a)

It looks like the [xtick labels](https://github.com/scikit-hep/mplhep/blob/c5664ba8858ace55ae6bfb96e9f942d1ae659e93/src/mplhep/plot.py#L564) are using the wrong values in `flow_bins` - modifying the indices used produces this plot:
![55aa4c23-f8cf-4a87-948c-89ca288ad1ba](https://github.com/scikit-hep/mplhep/assets/47793576/1797520e-a426-47bb-a717-05fb2dbbb300)

This looks nice if the bin centers line up with the x-ticks, however when this isn't the case you get:
![8e40351b-21e1-47c8-b912-1fd1977ef9d1](https://github.com/scikit-hep/mplhep/assets/47793576/8f0fc0d5-3b78-4928-9eff-6f9f3d457fe8)

This seems confusing - one possible solution is to replace any existing ticks in the flow bins with the flow bin center and label that. Doing this results in:
![b98f4458-ac8a-40de-b9ef-9d73956c8aea](https://github.com/scikit-hep/mplhep/assets/47793576/4c2b6383-27c4-4f61-9e4f-147b992e1923)

I can make a PR with these changes if wanted.

atownse2:
Also, when the axes passed to `histplot` have a shared x-axis, the `xticklabels` are sometimes empty, breaking the current code. After messing around for a bit I can get plots like this:
```python
h = hist.Hist(
    hist.axis.Variable([0.1, 0.4, 0.5, 0.9], underflow=True, overflow=True)
    ).fill(1*np.random.rand(1000))

fig, ax = plt.subplots(2, 1, sharex=True, gridspec_kw={'hspace': 0})

hep.histplot(h, flow='show', label='h', ax=ax[0])
hep.cms.label(ax=ax[0], data=False, year=2016)

# Draw random data on bottom plot
[...]

PR #502, opened 2024-06-03
fix: flow bins xticklabel correction

atownse2:
The issues that this PR fixes are described in #501. This includes:
- Use the correct flow bin index when drawing overflow `xticklabels`
- Draw flow `xticklabel` on overflow bin center instead of using default `xticks`
- Add support for shared axes: Draw overflow bin marker and dashed line on all shared axes, excluding the marker on top of top axis.

Here are some plots I can make with this PR:
5cfac6c0-34c9-41c2-b7b9-05105b80e6d1

9b02ff13-326c-4a9c-bf1d-975d92816e04


atownse2:
Also, to get the dashed lines and diamonds to draw correctly I needed to draw them on both the top and bottom of each axis. The diamond on the top of the top axis got in the way of labels so I removed it but I left the dashed line because I liked it. I'm not sure if this is the best choice in general.

src/mplhep/plot.py
```diff
@@ -550,68 +550,126 @@ def iterable_not_string(arg):
...
+
+            # Don't draw markers on the top of the top axis
+            top_axis = max(shared_axes, key=lambda a: a.get_position().y0)
+
+            # Draw on all shared axes
+            for _ax in shared_axes:
+                for h in [0, 1]:
+                    _ax.set_xticks(xticks)
+                    _ax.set_xticklabels(xticklabels)
+
```
andrzejnovak, line 603: If you move the skip condition up here, wouldn't that prevent drawing the dashed line on top?

andrzejnovak:
Hey, thanks for the PR and sorry for the delay in reviewing this. This looks mostly good to me, but I am not sure I understand why the dashed line needs to appear on top of the plot as well.

atownse2:
Hi, yes the dashed lines don't need to appear on the top. I thought that it made the overflow bin more clear. What I will do is only include the fixes in this PR and then I will make another PR with some drawing changes that I like.

andrzejnovak:
Thanks, so this now needs to update the reference images. Check https://github.com/scikit-hep/mplhep/blob/master/CONTRIBUTING.md and let me know if you have a problem getting the pytest generation running. N.b. the images generated by pytest might register as "new" in git even if they didn't change, please don't add those to the commit to limit the repo size.

andrzejnovak:
Thanks a lot @atownse2! I'll mint a new release with the fix shortly.

PR #506, opened 2024-06-11
fix: remove precision specifier for flow bin labels

atownse2:
I modified the flow bin labels to use the exponential formatter because I wanted to remove trailing zeros, however, it will use the specified precision to set the point at which it switches from float to exponential notation.

The code as is produces plots like this:
8e1c6589-6967-45b9-a49c-88d359f0daf4

My suggestion is to remove the precision specifier which will by default use exponential notation after e-4 or e+5. This switch over is shown here:
cd103d1f-0a72-4fda-9105-d96a1ead5c54
aa1cd4f7-aca1-4292-9ca6-1500fce3d624

It looks like the default is equivalent to setting the precision to 6, that is `.6g` = `g` as far as I can tell. This formatter doesn't seem to differentiate between the number of significant digits it keeps and the point at which it switches over from the float to exponential representation.

The original formatting that `mplhep` used was whatever the python f strings use by default. This shows all the decimals and truncates all trailing zeros except one. It will switch to exponential notation after e-4 but I'm not sure it ever switches over on the high side. For example:
a09e1d2c-ebbb-450c-ab4b-59a33d8eab8d

### DIFF
--- a/src/mplhep/plot.py
+++ b/src/mplhep/plot.py
@@ -551,8 +551,8 @@ def histplot(
             )
 
     elif flow == "show":
-        underflow_xticklabel = f"<{flow_bins[1]:.2g}"
-        overflow_xticklabel = f">{flow_bins[-2]:.2g}"
+        underflow_xticklabel = f"<{flow_bins[1]:g}"
+        overflow_xticklabel = f">{flow_bins[-2]:g}"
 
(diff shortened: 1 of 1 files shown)


andrzejnovak:
That's a good catch, thanks!

PR #510, opened 2024-08-06
fix: make left label read COM energy

Benkendorfer:
Previously, only right-side experimental labels actually read the COM energy. This commit makes the left-side labels do the same. The default remains at 13 TeV if no COM energy is supplied.

This was a problem with ATLAS labels for LHC Run 3, since the COM energy is now 13.6. We probably want the flexibility to use this package at various COM energies, so I did not just update the text from 13 to 13.6.

### DIFF
--- a/src/mplhep/label.py
+++ b/src/mplhep/label.py
@@ -414,12 +414,18 @@ def exp_label(
             units="inches",
             fig=ax.figure,
         )
+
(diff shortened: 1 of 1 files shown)


andrzejnovak:
Thanks for the fix! Sorry for the delay in merging, I was on holiday.

PR #514, opened 2024-08-28
Fix docstring for histtype

cverstege:


andrzejnovak:
Whoops. Nice catch. Thanks!

PR #516, opened 2024-09-11
feat: add function to merge handles for identical labels in legend

cverstege:
This function is useful, when using multiple plotting functions and one only wants to use a single label. This function combines all the handles for identical labels into a single handle.
I think this comes in handy. e.g. when doing a step plot for a main line and a band around for the uncertainty.

I'm open for feedback.


andrzejnovak:
Hi @cverstege , this looks ok for me and we already collect some "random" helpers. For posterity, could you add a couple of snippets of how you use it as a comment to this PR?

cverstege:
Sure. Here's one example:
```python
import mplhep as hep
import numpy as np


if __name__=="__main__":
    s = np.random.normal(size=1000)
    hist, bins = np.histogram(s)
    error = np.sqrt(hist)

    hep.style.use("CMS")
    fig, (ax1, ax2) = plt.subplots(2, 1)
    for ax in (ax1, ax2):
        hep.histplot(hist, bins=bins, yerr=error, histtype="band", label="same label", ax=ax, zorder=-1)
        hep.histplot(hist, bins=bins, histtype="step", color="black", label="same label", ax=ax, zorder=0)
    ax1.legend(loc="upper right")
    handles, labels = ax2.get_legend_handles_labels()
    handles, labels = hep.merge_legend_handles_labels(handles, labels)
    ax2.legend(handles, labels, loc="upper right")
    plt.savefig("test.pdf")
```

cverstege:
And here the plot for reference. The upper one is not using the new function, while the lower one has the combined legend entry.
test

PR #517, opened 2024-09-14
build: Remove 'toml' extra from setuptools_scm build requirement

matthewfeickert:
* The toml extra from setuptools-scm is now empty and exists only for legacy purposes and so can be removed.
* The modern setuptools-scm docs recommend a lower bound of setuptools v64 and setuptools-scm v8.
   - c.f. https://setuptools-scm.readthedocs.io/en/v8.1.0/


andrzejnovak:
Thanks!

PR #522, opened 2024-10-07
chore: modernize to hatchling backend

henryiii:
This drops setup.py, setup.cfg, MANIFEST.in, and uses hatching, which breaks a lot less and is faster than setuptools. The modern config also means uv's fancier features work out of the box. And in the rare case you see an error message, it's much better than the deep nesting of setuptools -> distutils.


setup.cfg
```diff
@@ -1,45 +0,0 @@
-[metadata]
-name = mplhep
-description = Matplotlib styles for HEP
-long_description = file: README.md
-long_description_content_type = text/markdown
-url = https://github.com/scikit-hep/mplhep
-author = andrzejnovak
-author_email = "novak5andrzej@gmail.com"
```
henryiii, line 8: This was a typo, btw - the quotes actually get stored in the metadata!

andrzejnovak:
Thanks!

henryiii:
NP, will be working on adding 3.13 next.

PR #523, opened 2024-10-07
ci: add 3.13 tests and classifier

henryiii:
Can we change the default branch to `main`? It's currently some old back port branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         runs-on: [ ubuntu-latest ]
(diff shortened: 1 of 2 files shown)


henryiii:
Any idea what's pulling pyarrow? Ahh, papermill is.

henryiii:
Just removed Papermill since it was commented out and unused.

jonas-eschle:
> Can we change the default branch to `main`? It's currently some old back port branch.

Yes, this was a bit buggy and wouldn't switch to main, but done now.

Looks all got to me, thanks a lot @henryiii

jonas-eschle:
@henryiii good point, done

PR #527, opened 2024-10-15
Correct ax.legend() order for stack hists

0ctagon:
Part of the [merging project](https://github.com/orgs/scikit-hep/projects/21).

### DIFF
--- a/src/mplhep/plot.py
+++ b/src/mplhep/plot.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import collections.abc
 import inspect
+import itertools
 import logging
 from collections import OrderedDict
 from typing import TYPE_CHECKING, Any, NamedTuple, Union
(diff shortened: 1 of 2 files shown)


0ctagon:
For the failing tests:
- `histplot_real`: now has a correct legend order
- `histplot_kwargs`: now has a correct legend order
- `histplot_stack`: plotting order has changed which modified a few pixels, but no differences

tests/test_mock.py
```diff
@@ -41,13 +41,14 @@ def test_simple(mock_matplotlib):
     bins = [0, 1, 2, 3]
     hep.histplot(h, bins, yerr=True, label="X", ax=ax)
 
-    assert len(ax.mock_calls) == 12
+    assert len(ax.mock_calls) == 13
```
0ctagon, line 45: Size is increased by the number of `ax.has_data()` called.

tests/test_mock.py
```diff
@@ -41,13 +41,14 @@ def test_simple(mock_matplotlib):
...
 
     ax.stairs.assert_called_once_with(
         values=approx([1.0, 3.0, 2.0]),
         edges=approx([0.0, 1.0, 2.0, 3.0]),
         baseline=0,
         label=None,
+        color=plt.rcParams["axes.prop_cycle"].by_key()["color"][-1],
```
0ctagon, line 52: Here the color cycle is not resetting to the original order as `ax.has_data()` (which, if false reset the color cycle) here calls `mock.has_data()` which is always true.

0ctagon:
Ready for review @andrzejnovak @jonas-eschle

src/mplhep/plot.py
```diff
@@ -409,6 +409,20 @@ def iterable_not_string(arg):
     ##########
     # Plotting
     return_artists: list[StairsArtists | ErrorBarArtists] = []
+
```
andrzejnovak, line 412:
```suggestion
# customize color cycle assignment when stacking to match legend
```

src/mplhep/plot.py
```diff
@@ -409,6 +409,20 @@ def iterable_not_string(arg):
     ##########
     # Plotting
     return_artists: list[StairsArtists | ErrorBarArtists] = []
+# customize color cycle assignment when stacking to match legend
```
andrzejnovak, line 412:
```suggestion
    # customize color cycle assignment when stacking to match legend
```

PR #528, opened 2024-10-17
feat: add plothist style

andrzejnovak:
@0ctagon @cyrraz @jonas-eschle


jonas-eschle:
Thanks for going forward with this! About the style name, isn't this maybe better to be named `BelleII` or similar?

andrzejnovak:
Yes, I would be in favour of naming it "Belle" which would be consistent with the rest of the styles and I assume safe wrt Belle vs Belle2 :) but I'd let @0ctagon and @cyrraz opine on that

0ctagon:
I think we need:
- The default style of plothist for the examples in the documentation, which we would call `plothist` for legacy or `default`, `base`, `masterclass`...
- The default Belle II style, which by construction would be very similar to the current plothist style, that we would call `Belle2`. To push this into mplhep, I think we would need approval from the Belle II publication committee (and they might change the style a bit from plothist). I will look into this.
- I have no idea about the Belle plot publishing guidelines, I will also look into this so we can add it.

andrzejnovak:
Well, let's put that in for now and we can revisit it / include dedicated Belle style later

PR #529, opened 2024-10-20
feat: added bar type + correct legend for step type

0ctagon:
Part of the [merging project](https://github.com/orgs/scikit-hep/projects/21), linked to [this PR in plothist](https://github.com/cyrraz/plothist/pull/177).

This PR adds:
1. Two new `histtype`: `bar` and `barstep`, which behave like matplotlib: if one histogram is provided `bar` and `barstep` are equivalent to `fill` and `step`. If multiple are provided, the bars or step are arranged side by side. The `barstep` can also display y error bars (see [tests/baseline/test_histplot_types.png](https://github.com/scikit-hep/mplhep/compare/main...0ctagon:barplot_histlegend?expand=1#diff-f8bd6d024d8c38a32538f9b4c57cd902064f8176b262d516f5965b7a481c1803) and [tests/baseline/test_histplot_bar.png](https://github.com/scikit-hep/mplhep/compare/main...0ctagon:barplot_histlegend?expand=1#diff-905e26888b075fcb4bc13ba946bc89b8dd47fa547e3fda094033b2706f43f862))
2. The `step` type has now the same representation in the legend as in matplotlib.
3. By default, `yerr` is not shown for `step` type.

@andrzejnovak @jonas-eschle


andrzejnovak:
This is great! I think we don't want to change the default behaviour of `yerr` for `step`. We had a discussion on the topic in the past and iirc the conclusion was that we want to let the user "know" right away if the object has that kind of information.

0ctagon:
I understand the reason and I reverted back to how it was before.
As matplotlib doesn't really support step with errorbar, in plothist we would represent them with what corresponds to the `band` type in mplhep. I think we will need a good section in the doc on how the errors are handled and displayed (in the current plothist doc, everything is boost-histogram so everything has errors already, which overloads a bit all the plots with yerr=True by default).

0ctagon:
If necessary we could add an argument to change the legend to a rectangle, but lets get the request first.

PR #531, opened 2024-10-23
fix: make fontsize same for experiment label and secondary label for ATLAS

diptaparna:
The font sizes for "ATLAS" and "Simulation/Internal/Preliminary" needs to be the same as per the latest ATLAS plot style.
However, in `mplhep`, the font size for "ATLAS" was intentionally 1.3x the font size of the secondary label.
This pull request fixes this issue.


ntadej:
I confirm this from the ATLAS side.

.gitignore
```diff
@@ -49,6 +49,8 @@ coverage.xml
 *.cover
 .hypothesis/
 .pytest_cache/
+flow_th1.root
+result_images/
```
matthewfeickert, line 53: Why are these needed?

diptaparna: They get created when you run pytest.

diptaparna:
Can we please merge this?

andrzejnovak:
Ok, thanks for the fix! Since I am not ATLAS, I was waiting for some contacts to corroborate the change.

andrzejnovak:
The release should appear on pypi shortly @diptaparna

matthewfeickert:
Relevant details from the [Style guide for ATLAS Plots](https://twiki.cern.ch/twiki/bin/view/AtlasProtected/PubComPlotStyle#ATLAS_labels):

> ## ATLAS labels
> 
> Use the ATLAS labels if and only if your plot deserves them. Rules that govern the use of the ATLAS labels can be found in the [Physics Policy Document PDF](https://twiki.cern.ch/twiki/pub/AtlasProtected/AtlasPolicyDocuments/Physics_Policy.pdf). If your plot has not been approved by ATLAS to be shown outside the Collaboration, it should not have an "ATLAS" or "ATLAS Preliminary" label.
> 
> 
> * The _ATLAS_ name **must** be draw in Helvetica, all upper case, bold face, slanted. The ATLAS Style package provides a macro to do it.
> * Any additional text after the _ATLAS_ name (e.g. "Preliminary") **must** be drawn in Helvetica normal typeface (thus, neither bold nor slanted), with first letter capitalized, in the same font size of the _ATLAS_ name. The ATLAS Style has an example on how to properly do it.
> 

This information would have been useful to include in the commit message itself.

PR #547, opened 2025-02-24
feat: hist2dplot labels options

rkansal47:
This adds arguments for fontsize, rounding, and color for hist2dplot. With these, I'll be able switch over all my 2d histogram plotting to mplhep :).


andrzejnovak:
Thanks!

Issue #550, opened 2025-03-06
Stacked, filled hists do not sort


rpsimeon34:
When using `histplot` on a list of `hist` histograms, `sort="yield"` has no effect when also using the options `stack=True` and `histtype="fill"`. The histograms are stacked in the order they are given, regardless of whether `sort` is unset, "yield", or "yield_r". Interestingly, switching "yield" to "yield_r" _does_ reverse the colors, but the stacking/plotting order is still unchanged.

As a potential hint, this does not happen when `histtype` is unset in v0.3.57. However, in v0.3.55, this issue occurs even for `histtype` unset.

Versions
--------

`numpy`: v.1.26.4
`mplhep`: v0.3.57
`hist`: v2.8.0

Reproducer
------------

```
from hist import axis, Hist
import mplhep as hep
import numpy as np

hist_ax = axis.Regular(10,0.0,1.0,name="foo",label="Axis label")
hist1 = Hist(hist_ax,name="Events")
hist2 = Hist(hist_ax,name="Events")
hist3 = Hist(hist_ax,name="Events")

hist1.fill(foo=np.random.rand(100)) #Biggest yield
hist2.fill(foo=np.random.rand(10)) #Smallest yield
hist3.fill(foo=np.random.rand(30))

hep.histplot([hist1,hist2,hist3],stack=True,histtype="fill",sort="yield")
```
As described, in this reproducer if I switch `sort` to "yield_r" I get the same (unsorted) plot, but with the colors reversed.

Please let me know if I can provide any additional info. Thank you!

andrzejnovak:
Thanks a lot for the report. There was a bit of a refactor not long ago and I guess we didn't have this in tests. Will take a look.

andrzejnovak:
This should be fixed now in https://github.com/scikit-hep/mplhep/releases/tag/v0.3.58 @rpsimeon34

rpsimeon34:
That works for me, thank you!

PR #551, opened 2025-03-07
fix: correct sorting

andrzejnovak:


andrzejnovak:
Fixes https://github.com/scikit-hep/mplhep/issues/550

PR #574, opened 2025-06-13
feat!: updating ATLAS color palette to match collboration recommendations

zlmarshall:
After an internal working group to discuss color palette options and color-vision deficiency of various kinds, these are the proposed colors that the experiment will adopt.

The plan is to announce this next week, and so it'd be great to have this in place by then.

Thanks,
Zach

### DIFF
--- a/src/mplhep/styles/atlas.py
+++ b/src/mplhep/styles/atlas.py
@@ -178,8 +178,10 @@ ATLASTex = {k: v for k, v in ATLASTex.items() if k in mpl.rcParams}
 ATLAS = ATLAS2.copy()
 
 # Add warnings for the older sets
-color_msg = ("This ATLAS style uses a color sequence not preferred by the collaboration. "
-             "Please switch to 'ATLAS2' or 'ATLAS' style. ATLAS members can see more at "
-             "https://twiki.cern.ch/twiki/bin/view/AtlasProtected/PubComPlotStyle#Colors")
+color_msg = (
+    "This ATLAS style uses a color sequence not preferred by the collaboration. "
+    "Please switch to 'ATLAS2' or 'ATLAS' style. ATLAS members can see more at "
+    "https://twiki.cern.ch/twiki/bin/view/AtlasProtected/PubComPlotStyle#Colors"
+)
 ATLAS1 = deprecated_dict(ATLAS1, message=color_msg, warning=FutureWarning)
 ATLASAlt = deprecated_dict(ATLASAlt, message=color_msg, warning=FutureWarning)
--- a/src/mplhep/styles/atlas.py
+++ b/src/mplhep/styles/atlas.py
@@ -24,13 +24,18 @@ color_sequence1 = [
 
 # Color wheel based on internal discussions of optimal
 # colors for visibility, accounting for color vision deficiency
+# The recommendation for signal is ROOT.kRed aka '#ff0000'; for
+# plots with no signal, '#ff0000' should be used as the final
+# color in the standard palette / color wheel.
+# For large signals, ROOT.kWhite / '#ffffff' is also an option
 color_sequence2 = [
-    "#e69f00",
     "#56b4e9",
-    "#009e73",
+    "#e69f00",
     "#f0e442",
-    "#0072b2",
+    "#009e73",
     "#cc79a7",
+    "#0072b2",
+    "#ff0000",
 ]
 
 _base = {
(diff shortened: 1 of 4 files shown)


src/mplhep/styles/atlas.py
```diff
@@ -5,18 +5,15 @@
...
-    "#e76300",
-    "#b9ac70",
-    "#717581",
-    "#92dadd",
+    "#e69f00",
+    "#56b4e9",
+    "#009e73",
+    "#f0e442",
+    "#0072b2",
+    "#cc79a7",
```
matthewfeickert, line 16: @zlmarshall Is there any external reference here (e.g. how the Petroff paper was linked before), or are these just "made up" by ATLAS?

matthewfeickert (commented):
Wow, that's disappointing. This is a real missed opportunity by ATLAS.

I can also say that this is a far worse color palette for me, a red green colorblind person. So big :-1: on this decision by ATLAS Pub Comm.

src/mplhep/styles/atlas.py
alexander-held: Maybe I just need to wait until next week but this comes as quite a surprise for a variety of reasons.

kratsg: @zlmarshall I can't see these color options very well myself. The older one is better than this one, so I'm a bit confused by how this was chosen if it's not part of the survey?

matthewfeickert:
> Oh, it's weird we don't have a test image for this.

We can add one.

jonas-eschle:
@zlmarshall is this more of a "no plot with other colors will ever be published again by ATLAS" or more of a recommendation that people should switch to? (I'll refrain from commenting about the choices, others have done that)

If simply the colors change, this is a breaking change, is this necessary?

Did you consider, in order not to break existing plots and maybe still give some choice, to introduce a naming scheme, as for example done for the [LHCb styles `LHCb1` and `LHCb2`, with `LHCb` being the moving style](https://mplhep.readthedocs.io/en/latest/gallery/styles.html#lhcb). LHCb2 is newer and better.

Making this the ATLAS2 style? Getting feedback and move ATLAS to the ATLAS2 style while preserving the current one as ATLAS1?

zlmarshall:
Ok, mercifully things are passing again. I suggest hanging on here, though, as I've found an inconsistency in the recommendation between what I was provided and what's been written down. Stay tuned...

Thanks,
Zach

zlmarshall:
Hi @andrzejnovak ,

the discussion within the collaboration made some useful points that are being reconsidered. Please give us a couple days. I’m happy to resolve any conflicts later in case you have overlapping MRs that need to go in. 

Thanks,
Zach

zlmarshall:
Hi @andrzejnovak , @jonas-eschle , @matthewfeickert ,

I've now implemented what I think will be our final (for now) recommended color palettes. Please have a look and let me know in case I've done anything incorrectly!

Cheers,
Zach

zlmarshall:
Hi @jonas-eschle ,

The documentation is now working, but those plots don't look anything like what our normal plots look like. We never have a filled stacked histogram for data, for example, which is the first thing on the top left. Data are points with error bars, in black. We very rarely show lines that are fits in plots. If you look at the vast majority of recent papers, you're looking at stacks of histograms. In some exceptional cases we have fit functions that are overlaid, but again the data are black points.

So: how much do you want me to massage this to make it into something like an ATLAS person would actually see? For reference, I'll attach the two plots we provide as examples for our members (which are behind the first link, which is ATLAS internal). I'd be happy to have @matthewfeickert weigh in there as well — I'm not an mplhep user, so I really don't know how much it matters if things in that example page actually match what we would expect to get out.

Best,
Zach

AtlasExampleOkabeIto

AtlasExampleArxiv

jonas-eschle:
> The documentation is now working
good, thanks!

> but those plots don't look anything like what our normal plots look like.

Yep, aware of this. We're currently merging quite some code from another repo including updating the docs, where we will add more examples for the stiles; they don't represent ATLAS (or CMS) plots well, agree.

> So: how much do you want me to massage this to make it into something like an ATLAS person would actually see? For reference, I'll attach the two plots we provide as examples for our members

I meant more a reference of the decision to use this colors, maybe there is some editorial board FAQ or anything like this. Just so that users can refer to the document in case asked. But maybe it's not a use-case.

Anyway, looks good to me (if you want/can add a reference to an EB doc or similar, free to do that)!

Any more changes? Otherwise we'll merge it (and possibly also create a new release of mplhep to distribute it, @andrzejnovak ?)

jonas-eschle:
Perfect! LGTM!

PR #576, opened 2025-06-30
test: cleanup matplotlib rc params before and after each test

jonas-eschle:
Adding a fixture that runs before and after each test. Ensures the parameters are reset.


andrzejnovak (approved):

jonas-eschle:
They're independent, but gonna fix them. Mostly about imports not at the top level, but in the tests, i.e. importing boost histogram, we can put this to the top, right?

PR #578, opened 2025-07-09
feat: add comparison plotter functions

0ctagon:
Add all the comparison functions from plothist.


0ctagon:
Will re-open it when tests are added.

PR #585, opened 2025-08-26
chore: update citation.cff, include Jonas, use general DOI

jonas-eschle:
Use DOI valid for all versions, adding myself


jonas-eschle:
@andrzejnovak FYI

PR #616, opened 2025-10-16
Import ParamSpec in compatibility module for typing

jonas-eschle:


.github/workflows/ci.yml
```diff
@@ -48,7 +48,13 @@ jobs:
           echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | sudo debconf-set-selections
           sudo apt-get install ttf-mscorefonts-installer
 
-      - name: Install package
```
jonas-eschle: @andrzejnovak added the check to make sure it install properly without the `all`, at least on a minimal level

PR #636, opened 2025-11-13
feat: drop Python 3.8

henryiii:
Python 3.9 is past EoL and already dropped by pytest and coverage. But since this still lists 3.8, that should probably be dropped first. :)

Also, there are no tests for 3.8 in CI! This means it could break at any time. It's much better to drop it to ensure old versions keep working and are not replaced by broken new versions.


andrzejnovak:
Thanks! Indeed, I've been meaning to get to it.

PR #651, opened 2025-11-17
chore: enable ALL for ruff

henryiii:
Enabling ALL, with explicit ignores.


andrzejnovak:
Looks like the latex test just needs `--benchmark-disable` https://github.com/scikit-hep/mplhep/actions/runs/19460534408/workflow?pr=651#L71 I assume "warning" was always showing up, but until some of the other patches it was not erroring out because it's harmless

PR #664, opened 2025-12-10
fix: add LHCb3 style 

### LINKED ISSUE #401: fix: mplhep LHCb2 style conflicts with subplots_adjust(hspace=0)
I want two arrange two axes with no space between, like the last figure [here](https://hist.readthedocs.io/en/latest/user-guide/notebooks/Histogram.html). I have this:
```
import mplhep
#mplhep.style.use("LHCb2")

fig,axs = plt.subplots(2,2)
plt.subplots_adjust(hspace=0)
plt.show()
```
where you see I get what I want (vertically) unless I uncomment the mplhep.style... line, which seems to block the behaviour I want. Does anyone know how I can use the LHCb style and achieve this axis-compression?

ghost:
Resolves #401

As requested, this PR adds a new style `LHCb3` which is functionally identical to `LHCb2` but with `figure.autolayout` removed. 

The default `LHCb` alias now points to `LHCb3`.

Verified with `pytest` in `tests/test_lhcb3.py`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.7
+  rev: v0.14.8
   hooks:
     # Run the linter.
     - id: ruff-check


tests/test_lhcb3.py
```diff
@@ -0,0 +1,21 @@
...
+import pytest
+
+import mplhep as hep
+
+
+@pytest.mark.mpl_image_compare(style="default", remove_text=True)
+def test_lhcb3_simple():
+    """Test that LHCb3 loads and renders without error."""
+    hep.style.use("LHCb3")
+    fig, ax = plt.subplots()
```
andrzejnovak: Maybe the test should use actual 2 axes layout to verify the behaviour no?

andrzejnovak (changes requested):

PR #650, opened 2025-11-17
feat(LHCb)!: Termes experiment label; tighter margins in LHCb2

jonas-eschle:
## Summary

Retarget of the original LHCb3 draft into direct updates of **LHCb2** (per discussion — a separate style wasn't warranted once `figure.autolayout` was ruled out, see #401):

- **`exp_lhcb` label/text default to `Tex Gyre Termes`** (bundled in `mplhep_data`) instead of the hard-coded `Times New Roman`. This was the last implicit dependency on the non-redistributable MS core fonts — LHCb2 body text has been Termes since #330. Termes is metric-compatible with Times (it descends from the same URW Nimbus Roman that ROOT uses for font 132), so layouts are unchanged.
- **LHCb2 `axes.ymargin: 0.0`** — matches the already-set `xmargin: 0.0`, from the LHCb3 draft.
- **LHCb2 `savefig.pad_inches: 0.3`** (was 0.1), from the LHCb3 draft.
- Not adopted from the draft: `figure.autolayout: true` (breaks `subplots_adjust(hspace=0)`, #401) and the LaTeX preamble trim (dropping `amsmath` could break user math in TeX mode).

Baselines regenerated for the five affected tests (`test_style_lhcb`, `test_style_lhcb2`, `test_style_lhcbtex2`, `test_style_ratio_plot_lhcb`, `test_style_ratio_plot_lhcb2`). `LHCbTex2` inherits the margin changes, hence the lhcbtex2 baseline; `test_style_lhcbtex1` is unaffected, confirming the label font change is inert under `usetex`.

## Follow-up

With no test depending on Times New Roman anymore except `LHCb1` body text, the `ttf-mscorefonts-installer` steps in CI can likely be dropped (they're a recurring source of flaky failures — SourceForge 522s — see #751).

## Test plan

- [x] Full suite passes locally at zero tolerance (374 passed)
- [ ] CI green on all platforms

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Eb2vYvYgC8rW2hCTzeK3ru

PR #674, opened 2025-12-31
fix: LHCb2 style step histogram errorbar width; add color scheme to tests

### LINKED ISSUE #401: fix: mplhep LHCb2 style conflicts with subplots_adjust(hspace=0)

### LINKED ISSUE #613: fix: errorbar width for ATLAS/LHCb styles
As we can now see in the docs both ATLAS and LHCb styles exhibit a strange behaviour for errorbars with `histtype='step'` where these are imo  too thick. 

Image

At first I wondered if this is for data-like plots, but `histtype='errorbar'` doesn't have this behaviour 

Image

@jonas-eschle @matthewfeickert can you comment on LHCb/ATLAS sides if this is intended behaviour?

andrzejnovak:
## Summary

- **Add LHCb3 style** with `figure.autolayout` disabled, allowing `subplots_adjust(hspace=0)` to work properly (fixes #401)
- **Fix errorbar linewidth mismatch** in step histograms by matching `elinewidth` to the step edge width (fixes #613)

This PR incorporates and extends the work from #650 by @jonas-eschle.

### LHCb3 Style
- New LHCb3 style based on LHCb2 with improvements
- Disables `figure.autolayout` to allow manual subplot adjustment
- Users who need autolayout can still enable it manually

### Errorbar Width Fix
When using `histtype='step'` with errors, the step edges use `patch.linewidth` while errorbars previously used `lines.linewidth`. In styles like LHCb where these differ (2pt vs 3.3pt), this caused visual inconsistency.

Now errorbars automatically match the step edge width:
- Default style: both 1.5pt
- LHCb2 style: both 2.0pt

## Test plan
- [x] All histplot tests pass with regenerated baselines
- [x] Visual verification shows matching linewidths
- [x] Verified fix works with both default and LHCb styles

Closes #401
Closes #613
Supersedes #650


src/mplhep/styles/__init__.py
```diff
@@ -17,7 +17,7 @@
 from .atlas import ATLAS, ATLAS1, ATLAS2, ATLASAlt, ATLASTex
 from .cms import CMS, ROOT, CMSTex, ROOTTex
 from .dune import DUNE, DUNE1, DUNETex, DUNETex1
-from .lhcb import LHCb, LHCb1, LHCb2, LHCbTex, LHCbTex1, LHCbTex2
+from .lhcb import LHCb, LHCb1, LHCb2, LHCb3, LHCbTex, LHCbTex1, LHCbTex2, LHCbTex3
```
0ctagon: Is it clear for the users, without looking at the code, the differences between the styles and which one they should use? Maybe having a documentation page with *all* the available styles to compare them would be nice. Also adding small guidelines (unless this is internal to LHC review process).

src/mplhep/styles/lhcb.py
```diff
@@ -238,6 +238,119 @@
     "pgf.rcfonts": False,
 }
 
+colors3 = [
```
0ctagon: What's the use of colors3?

PR #711, opened 2026-03-11
fix: Correct label positioning across figure sizes and DPIs

andrzejnovak:
This PR addresses issue #709 where the positioning of `mh.<exp>.label` (such as the gap between "CMS" and "Preliminary") was inconsistent across different figure sizes and aspect ratios. 

The root cause was traced to how font sizes (which are provided in typography points, 1 point = 1/72 inch) were being converted into axis fraction units. The code was incorrectly dividing the point size by `ax.figure.dpi` (effectively treating points as pixels) instead of by `72`. 

This has been fixed in `_fontsize_axis` and `append_text` within `src/mplhep/label.py` by properly converting points to inches via `fontsize / 72`. Test baselines generated via `pytest-mpl` have been updated to reflect this corrected spacing.

---
*PR created automatically by Jules for task [14182033389645325178](https://jules.google.com/task/14182033389645325178) started by @andrzejnovak*


andrzejnovak:
Force-pushed a revised version of the fix. Same correct math (points→inches, no `/dpi`), but the calibration constants are retuned so the rendered output at the default DPI of 100 is **byte-identical** to current `main` — no baseline regeneration needed.

### What changed vs the original commit

| | Original commit | This commit |
| --- | --- | --- |
| `_fontsize_axis` math | `(fontsize/72) / ax_height` | same |
| `text_width_corr` / `text_height_corr` math | `(fontsize/72) / FACTOR / ax_width` | same |
| `FONT_WIDTH_CORRECTION_FACTOR` | 3 | `3 × 100/72` |
| `FONT_HEIGHT_CORRECTION_FACTOR` | 10 | `10 × 100/72` |
| `_inside_pad` multiplier | `_fontsize_axis × 100` | `_fontsize_axis × 72` |
| Baseline images regenerated | 18 | 0 |

### Why retune

Without the calibration constants, the dimensionally-correct formula makes every label gap ~38% wider at default DPI (factor 100/72 ≈ 1.388). That would silently shift CMS/ATLAS/etc. labels everywhere and was the reason 18 baselines moved in the original PR. Pre-tuning the correction factors by `100/72` cancels that exactly at the calibration DPI.

### Verification

- At `dpi=100` (legacy default), pixel gap matches main exactly: **8.67 px** on every figsize tested (10×10, 12×5, 16×4, 20×3, 5×12, 6×6).
- Pixel gap scales linearly with DPI now — constant **0.087 inches** across dpi=50/100/200/300 (vs current main which keeps a fixed 8.67px regardless of DPI, so the inch-gap shrinks at print resolution).
- Full test suite: **323 passed, 0 failed** against the existing baselines.

So the fix is now strictly a DPI-correctness improvement that's invisible at the default DPI and only kicks in when users render at non-100 DPI (e.g. `savefig(..., dpi=300)` for publication).

PR #728, opened 2026-06-11
fix: better com defaults 

GiorgosChr:
Added `None` option for experiment label. Removes com text entirely.
Other units can be used to replace TeV.


GiorgosChr:
I think my changes pass all the tests, it should be okay to merge.

andrzejnovak:
Thanks @GiorgosChr  will mint a new release

Issue #731, opened 2026-06-12
mplhep review from Claude Fable


henryiii:
:robot: _AI text below_ :robot:

All findings verified. Here's the consolidated review.

# mplhep review

The codebase is in good shape overall — modern typing, active ruff `ALL`, current CI actions. But the review found **several real, reproduced bugs** (two of which silently plot wrong statistics), a substantial amount of removable dead code (~1,000+ lines), and one significant performance hot spot in the y-rescaling pipeline.

## Bugs (verified, ordered by severity)

**1. Wrong Poisson error bars on empty bins of weighted histograms** — `src/mplhep/error_estimation.py:44`. The "nearest nonzero bin" logic uses `np.sum([...])`, which collapses the list of distance matrices to a *scalar*, so `.argmin(axis=0)` is always 0 and every empty bin inherits the scale of the *first* nonzero bin instead of the nearest. The original coffea code used builtin `sum(...)` (element-wise over the list); this regressed in commit `f027c68` ("fix: mypy now passes"). For `sumw=[4,0,0,9]`, `sumw2=[8,0,0,9]` the scale comes out `[2,2,2,1]` instead of `[2,2,1,1]` — silently wrong plotted uncertainties.

**2. `histplot(..., sort="label")` silently drops histograms** — `src/mplhep/plot.py:495`. It sorts `np.argsort(label)` on the raw argument instead of the normalized `_labels` list. With a scalar `label="data"` (or `None`), `argsort` returns `[0]` and lines 509–511 then keep only one histogram. Also, an unrecognized sort string like `"foo"` leaves `order` unbound → `UnboundLocalError` instead of the intended `ValueError`.

**3. `hist(..., weights=w, yerr=False)` draws error bars anyway** — `src/mplhep/plot.py:234`. `yerr_arg = None if w2_arg is not None and isinstance(yerr, bool) else yerr` converts both `True` *and* `False` to `None`, and `yerr=None` + `w2` makes histplot compute and draw errors. Should be `... and yerr is True`.

**4. `set_ylow(ax, ylow=value)` ignores the value** — `src/mplhep/utils.py:289`. The branch taken when `ylow` is provided runs `ax.set_ylim(0, ...)` — hardcoded 0. Also affects `mpl_magic(ylow=...)`.

[...]

PR #732, opened 2026-06-12
fix: nearest-bin scale lookup in poisson_interval for weighted histograms

henryiii:

Part of #731

## The bug

In `poisson_interval` (`src/mplhep/error_estimation.py`), empty bins of a weighted histogram (`sumw == 0`) borrow the per-bin scale `sumw2/sumw` of the **nearest** nonzero bin. The nearest-bin search built a list of per-dimension squared-distance matrices (each of shape `(n_available, n_missing)`) and was supposed to sum them element-wise, then take `.argmin(axis=0)` to find, for each empty bin, the closest nonzero bin.

A regression in f027c68 ("fix: mypy now passes") changed the builtin `sum([...])` to `np.sum([...])`. Unlike builtin `sum`, `np.sum` of a list of arrays collapses everything to a **scalar**, so `.argmin(axis=0)` was always `0`. Every empty bin therefore inherited the scale of the **first** nonzero bin rather than the nearest one, producing incorrect Garwood/Poisson error bars.

### Numeric example

```
sumw  = [4, 0, 0, 9]
sumw2 = [8, 0, 0, 9]
```

Per-bin scale `sumw2/sumw` is `2` and `1` for the nonzero bins. Bin 1 is nearest bin 0 (scale 2); bin 2 is nearest bin 3 (scale 1), so the expected scale vector is `[2, 2, 1, 1]`. The buggy code produced `[2, 2, 2, 1]` — bin 2 wrongly got the first nonzero bin's scale.

## The fix

Restore element-wise summation in a mypy-clean way:

```python
nearest = np.sum(
    np.stack([
        np.subtract.outer(d, d0) ** 2
        for d, d0 in zip(available, missing, strict=True)
    ]),
    axis=0,
).argmin(axis=0)
```

This preserves the `(n_available, n_missing)` shape and the existing N-dimensional behavior (indices come from `np.where`/`np.nonzero` tuples).

## Tests

Added two plain unit tests in `tests/test_utils.py` (no image comparison, no tolerance changes):

- `test_poisson_interval_nearest_nonzero_scale` — the 1D example above; fails on `main`, passes with the fix.
- `test_poisson_interval_nearest_nonzero_scale_2d` — confirms the N-dimensional nearest-bin lookup keeps working for a 2D `sumw`.

Full suite: `285 passed, 74 skipped` (skips are Linux-only style tests). Lint clean.


andrzejnovak:
@copilot resolve conflicts

PR #734, opened 2026-06-12
fix: scale bar/barstep widths by bin width

henryiii:

## The bug

For `histtype="bar"` and `histtype="barstep"`, the bar width (`bar_bin_width`, default `0.8`) and the side-by-side `_shift` offsets were applied in **absolute data units**, which silently assumed unit-width bins.

With bin edges `[0, 10, 20, 30]` and 2 histograms, each bar came out `0.8 / 2 = 0.4` data units wide — only 4% of the 10-wide bin instead of the intended 40% (sliver bars). Conversely, with bins narrower than `0.8` data units, bars from adjacent bins overlapped. All existing test baselines happened to use unit-width bins, which masked the bug.

## The fix

`bar_bin_width` is interpreted as a **fraction of each bin's width** and scaled per-bin via `np.diff(edges)` (computed per-plottable, so flow bins are handled too). Both the `bar` and `barstep` paths, including the `barstep` errorbar offsets, are scaled.

- For unit-width bins, `np.diff(edges) == 1`, so `0.8 × 1 = 0.8` and the rendered output is **bit-identical** to before — all existing pytest-mpl baselines pass unchanged at their current (zero) tolerance.
- Variable-width bins now render each bar scaled to its own bin's width.

Also:
- Documented `bar_bin_width` in the `histplot` docstring (as a fraction of the bin width).
- Simplified the `bar_bin_width` plumbing: it is popped once before per-hist kwarg chunking instead of in each per-branch path.
- Updated usage in tests to the reviewed kwarg name `bar_bin_width`.

## Test

Added `test_histplot_bar_nonunit_bins`, a new pytest-mpl image-comparison test exercising both `bar` and `barstep` with wide bins (`[0,10,20,30]`) and variable-width bins (`[0,1,4,10,25]`) for 2+ histograms.

🤖 Generated with <a href="https://claude.com/claude-code">Claude Code</a>


andrzejnovak:
@copilot update this pr for the new kwarg to read `bar_bin_width` instead of proposed

PR #736, opened 2026-06-12
chore: remove dead code and trivial duplication

henryiii:

Part of #731

## Summary

This is a pure cleanup PR — no behaviour changes. All 283 tests pass (74 pre-existing skips for Linux-only and missing ROOT module). Mypy and prek lint both clean.

### What was removed / fixed

| # | Item | Status |
|---|------|--------|
| 1 | Delete `src/mplhep/plot_comp.py` (892 lines) | Done — zero imports anywhere |
| 2 | `_utils.py` `_hist_legend` — dead function | Done — deleted |
| 3 | `_utils.py` `_overlap(get_vertices=...)` — unused param | Done — removed param and tail branch |
| 4 | `_utils.py` duplicate `ax.scatter(marker="+")` in `_DEBUG_OVERLAP` | Done — removed one copy |
| 5 | `comp.py` broken `compare = __import__(__name__)` | Fixed — replaced with `sys.modules[__name__]` |
| 6 | `__init__.py` `"add_text"` listed twice in `__all__` | Done — removed duplicate |
| 7 | `_utils.py` `flow=="none"` branch identical to trailing `else` | Done — merged into `else` |
| 8 | `_utils.py` `__add__` redundant `else: added_variances = None` | Done — dropped |
| 9 | `plot.py` two consecutive identical `if model_type == "histograms":` | Done — merged |
| 10 | `label.py` `_construct_filename` re-defined inside loop | Done — hoisted out |
| 11 | `label.py` duplicate LaTeX wrapping blocks | Done — extracted `_apply_latex_font_commands()` helper |
| 12 | `_utils.py` edges extraction reimplements `_get_plottable_protocol_bins` | **Skipped** — behavior differs for categorical axes (`NotImplementedError` vs label return), safe to leave |
| 13 | `exp_cms.py` commented-out deprecation-example dead code | Done — deleted |

### Question for the maintainer (item 1)

`plot_comp.py` was never imported by `__init__.py` or any other module — it is only reachable via an explicit `import mplhep.plot_comp`. The module is also subtly broken: its `data_model` calls `comparison()` passing a `w2method=` kwarg that `comparison()` does not accept. No tests reference it. Given that `comp.py` already provides the same API (with flow/blind support added later), deletion seems correct — but if there are downstream users relying on `import mplhep.plot_comp` for backwards compatibility this should be noted in the changelog / a deprecation warning should have been emitted first. Please confirm deletion is intentional.


andrzejnovak:
@copilot resolve conflcits

Issue #753, opened 2026-07-23
Docs deploys from release tags never publish; only `main` deployments serve


andrzejnovak:
## Summary

Only GitHub Pages deployments originating from `main` actually serve. Deployments from any
other ref — including `v*` tags — are recorded as `success`, mark the previous deployment
`inactive`, and never reach the live site. This means **versioned docs deploys from release
tags have not been publishing.**

## Evidence

Observed on 2026-07-23 while re-deploying the 1.0/1.1/1.2 docs:

| Deployment | Ref | Result |
| --- | --- | --- |
| 07:25:43 | `main` | served (site content matched this build) |
| 07:30:42 | `v1.3.2` | `success`, never served |
| 10:19:48 | `tmp-publish-pages` | `success`, never served |
| 10:37:43 | `tmp-publish-pages` | `success`, never served |
| 11:23:53 | `main` | served |

The decisive check: the `v1.3.2` run ends with `mike set-default --push latest`, so if that
deployment were serving, the site root would redirect to `latest/`. It did not:

```
live root redirect      ->  url=dev/     # what the 07:25 main run set
published gh-pages root ->  url=latest/  # what the tag run committed
```

The `gh-pages` branch and the uploaded artifact were both verified correct at the time —
`1.0/guide_comparisons/index.html` was 1,026,097 bytes in the artifact while the live site
served 32,302,746 bytes. Cache was ruled out: `cf-cache-status: DYNAMIC`, and cache-busted
plus `Cache-Control: no-cache` requests returned the same stale bytes for 11+ minutes. A
`main` deployment then went live within ~4 minutes.

This is consistent with Pages being configured `build_type: workflow` with
`source: {branch: "main", path: "/"}`.

## Impact

- Docs for a new release are committed to `gh-pages` by `mike` but never published. The site
  keeps serving whatever the last push to `main` produced.
- Because the `main` path ends with `mike set-default --push dev`, the site root
  permanently redirects to `dev/`. `latest/` is reachable, just never the default —
  `mike set-default --push latest` only runs in the tag path, which never takes effect.

## Possible fixes

1. Have the tag path trigger a publish from `main` (e.g. a follow-up `workflow_dispatch` on
   `main` after the tag deploy commits to `gh-pages`).
2. Reconsider the Pages source configuration so non-`main` refs can publish.
3. Stop hardcoding `mike set-default --push dev` in the `main` path, so the default is not
   silently flipped away from `latest` on every merge.

[...]

PR #754, opened 2026-07-23
chore(docs): document lean clones; harden retro-deploy of old tags

andrzejnovak:
## Summary

Two unrelated docs-infrastructure fixes, both found while re-deploying the 1.0/1.1/1.2 documentation.

### Lean clones (`CONTRIBUTING.md`)

`gh-pages` holds the built docs for every released version — ~50 MB, against ~26 MB for all
code branches combined. A default `git clone` therefore ships the published docs site to
every contributor, and nothing in the development workflow needs it.

Documents a clone that stays lean **without** losing branch visibility, by excluding only
`gh-pages` with a negative refspec rather than narrowing to a single branch:

```bash
git clone --single-branch --branch main https://github.com/scikit-hep/mplhep.git
cd mplhep
git config remote.origin.fetch '+refs/heads/*:refs/remotes/origin/*'
git config --add remote.origin.fetch '^refs/heads/gh-pages'
git fetch
```

Measured on a real clone: **28 MB** of `.git` with every code branch visible to
`git branch -r`, so `git checkout some/branch` and PR review work normally. Also documents
fetching `gh-pages` on demand (28 MB → 71 MB) and slimming an existing clone
(69 MB → 27 MB). Every figure and command in the section was verified against actual clones.

One non-obvious gotcha is called out: `git fetch --prune` does **not** remove a
remote-tracking ref that falls outside the current refspec, so it has to be deleted
explicitly before `git gc` can reclaim anything.

### Retro-deploy of old tags (`ci-pages.yml`)

The `deploy_version` path had two problems, both hit when re-deploying v1.0.0's docs:

- It ran `mike deploy --update-aliases "$VERSION" latest` and `mike set-default --push latest`
  unconditionally, so re-deploying an **old** tag repointed the `latest` alias and the site
  default at it. Now gated behind a new `make_latest` input, defaulting to `false`.

- Old tags need an era-correct matplotlib. Building them against the current floor does not
  fail — it silently mis-renders. Pre-1.3 `label.py` assumed the `Text._get_layout` format
  change landed in mpl 3.12 when it actually landed in 3.11, so under a current matplotlib it
  takes the wrong branch and computes a bad descent, displacing experiment labels. This is
  exactly what left the 1.1 and 1.2 docs with overlapping labels. Adds a `matplotlib_pin`
  input, applied after the `--no-deps` reinstall, plus a line echoing the resolved versions.

Dispatch inputs also moved out of `${{ }}` interpolation into `env:`, and `set -euo pipefail`
added.

## Testing

The `matplotlib_pin` approach is what produced the current live 1.0/1.1/1.2 docs. The v1.0.0
rebuild was pixel-compared against the original published build and is visually identical
apart from the plugin's dpi default change.

Not covered here: docs deploys from release tags never actually publish — see #753.

### DIFF
--- a/.github/workflows/ci-pages.yml
+++ b/.github/workflows/ci-pages.yml
@@ -11,6 +11,15 @@ on:
         description: 'Delete a mike version (e.g. pr-683). Leave empty to skip.'
         required: false
         type: string
+      make_latest:
+        description: 'Point the "latest" alias and the site default at deploy_version. Leave off when re-deploying an OLD tag.'
+        required: false
+        default: false
+        type: boolean
+      matplotlib_pin:
+        description: 'Optional matplotlib constraint for the deploy_version build (e.g. "matplotlib>=3.10.8,<3.11"). Old tags predate the mpl>=3.11 Text._get_layout change and crash without this.'
+        required: false
+        type: string
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


andrzejnovak:
Closing in favour of a cleaner split: the CONTRIBUTING.md clone guidance and the AGENTS.md policy change go in a new PR. The `ci-pages.yml` retro-deploy hardening (`make_latest`, `matplotlib_pin`) is preserved on this branch and is better paired with the fix for #753.

PR #766, opened 2026-08-04
fix: correctly propagate line style in prop cycler for mpl 3.11

kpedro88:
## Description

Previously fixed in #704, broken in #738 (I think).

Not sure why the tests added for this in #704 still passed.

## Proposed commit message

```text
fix: correctly propagate line style in prop cycler for mpl 3.11
    
Assisted-by: claude-code:claude-opus-4-8
```

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/scikit-hep/mplhep/blob/main/CONTRIBUTING.md)
- [ ] Tests added or updated for the change if changing code
- [ ] Only genuinely modified baseline images are included

## AI assistance

See the [AI Policy](https://github.com/scikit-hep/mplhep/blob/main/CONTRIBUTING.md#ai-policy). Pick one:

- [ ] No AI was used, or only for translation/grammar.
- [x] AI was used. Tool and model: claude-code:claude-opus-4-8

  By checking this you confirm you have reviewed and understood every line, you
  will respond to review comments yourself, and the commit message credits the
  assistance with `Assisted-by: <harness>:<model>` — never `Co-authored-by:` or
  `Signed-off-by:`.


kpedro88:
The tests "fail" because the baseline indeed only has solid lines in the reference images.

(I do not understand why there are changes in assertion values in `test_mock.py`.)

andrzejnovak:
Thanks for catching this @kpedro88. There were quite a few overhauls so it's likely we simply missed when the baseline changed but wasn't supposed to. 

Anyway, can you point your claude to CONTRIBUTING.md and have it regenerate the diffing baselines? 

To the extent that I understand the mock tests, they simply count the number of relevant mpl fcn calls, which changes in your PR so the reference assert value needs to reflect that.

kpedro88:
I am working on making the updates following CONTRIBUTING.md. I notice that the instructions are incomplete: the tests will not run after `pip install` *unless* LaTeX is also installed, with sufficient fonts. (Otherwise, there are many failures with `RuntimeError: latex was not able to process the following string: b'lp'` caused by missing files/packages.)

andrzejnovak:
Good catch we should separate latex dependent tests

kpedro88:
I don't think the CD/dist failure is caused by this PR...

andrzejnovak:
Indeed, it was a dependency issue. Thanks a lot for noticing this and sending the fix!

